PR #70, opened 2026-04-24
feat(call-graph): external leaf nodes, edge typing, content-hash cache & PostToolUse hook

laurentftech:
## Motivation

The call graph was originally useful only for internal topology — hub functions, entry points, blast radius. But it silently dropped any call that couldn't be resolved to a project-internal function. In practice this meant the graph went dark exactly at the most interesting boundaries:

- \`fetch\` / \`axios.post\` / \`https.request\` — HTTP calls, where SSL errors, timeouts, and auth failures happen
- \`redis.get\` / \`pg.query\` — database calls, where latency and connection pool issues surface
- \`Promise.all\` / \`Array.map\` with callbacks — concurrency and data transformation shapes

When trying to debug an SSL error, \`trace_execution_path(callEmbeddingsApi, fetch)\` returned *no path found* — because \`fetch\` didn't exist in the graph. The trail went cold one hop before the problem.

### 1. External leaf nodes + ExternalKind classification

**Before:** two \`continue\` statements in \`resolveEdges()\` silently discarded any call that failed all four resolution strategies.

**After:** a synthetic \`FunctionNode\` is created (\`id: external::<name>\`, \`filePath: "external"\`, \`isExternal: true\`) and the edge is recorded with \`confidence: "external"\`. External nodes appear in \`get_subgraph\` and \`trace_execution_path\` but are excluded from structural statistics (hubs, entry points, refactor priorities) so they don't pollute the analysis.

Each external node is classified by \`externalKind\`:

| Kind | Examples |
|---|---|
| \`http\` | \`fetch\`, \`axios\`, \`requests.get\`, \`reqwest::get\`, \`Alamofire\` |
| \`database\` | \`pg\`, \`prisma\`, \`sqlalchemy\`, \`gorm\`, \`ActiveRecord\` |
| \`filesystem\` | \`fs\`, \`open\`, \`fopen\`, \`File\` |
| \`stdlib\` | \`Array.isArray\`, \`Math.floor\`, \`fmt.Println\`, \`System.out\`, \`os.path\` |
| \`unknown\` | anything else |

**Stdlib nodes are filtered from \`get_subgraph\` output** — both from the node list and from edges. \`Array.isArray\`, \`t.slice\`, \`response.json\`, \`body.slice\` are noise; they disappear automatically. Only semantically meaningful externals are shown.

Classifiers cover all 8 supported languages: TypeScript/JavaScript, Python, Go, Rust, Ruby, Java, C++, Swift.

\`\`\`
// Before
trace_execution_path(callEmbeddingsApi, fetch)  →  no path found
get_subgraph(callEmbeddingsApi, downstream)     →  0 downstream nodes

// After
callEmbeddingsApi → [external] fetch            →  1 hop  ✓  externalKind: "http"
(Array.isArray, t.slice, body.slice filtered out as stdlib)
\`\`\`

### 2. Edge typing (\`EdgeKind\` + \`CallType\`)

Edges previously only carried \`confidence\` (how the edge was *resolved*). There was no distinction between a direct call, an async call, a constructor, or a test exercising production code.

**\`EdgeKind\`** (\`calls | tested_by\`) captures the *relationship type*. A \`tested_by\` edge is emitted automatically for every \`calls\` edge whose caller lives in a test file — so \`get_test_coverage\` and \`analyze_impact\` can traverse test relationships without separate tooling.

**\`CallType\`** (\`direct | method | awaited | constructor\`) is detected from AST parent context: an \`await_expression\` parent → \`awaited\`, a \`new_expression\` node → \`constructor\`, a receiver (\`obj.method()\`) → \`method\`, fallback → \`direct\`.

### 3. Content-hash fingerprint cache

**Before:** \`isCacheFresh()\` returned true if \`llm-context.json\` was less than 1 hour old — meaning an unchanged repo re-analyzed every hour, and a changed repo could serve stale results for up to 59 minutes.

**After:** \`computeProjectFingerprint()\` walks source files and hashes their mtimes+sizes (stat-only, no file reads, ~5 ms on a 1 k-file project). The hash is written to \`fingerprint.json\` after every analysis. On the next invocation, the hash is recomputed and compared: identical → instant cache hit regardless of age; different → re-analysis immediately. Size is included alongside mtime so the fingerprint survives \`git checkout\` (which resets mtimes but preserves sizes).

### 4. PostToolUse hook (Claude Code) + agent-guard auto-analyze (OpenCode/KiloCode)

\`spec-gen setup --tools claude\` now writes a \`PostToolUse\` hook to \`.claude/settings.json\` that triggers \`spec-gen analyze\` in the background after every file edit, debounced with a 10 s lockfile.

\`spec-gen setup --tools opencode\` installs \`agent-guard.ts\` which does the same via the \`tool.execute.after\` plugin hook, with an in-memory 10 s debounce.

Combined with the fingerprint cache: after every file edit the graph re-runs in ~seconds (fingerprint diff detects changes instantly), and the MCP tools always have a fresh graph without the user ever typing \`analyze_codebase\`.

## Test plan

- [x] \`npm run build\` — no TypeScript errors
- [x] \`npm test\` — full unit suite green
- [x] \`get_subgraph(callEmbeddingsApi, downstream)\` — 2 nodes (\`callEmbeddingsApi\` + \`[external] fetch\`), 1 edge; stdlib nodes absent
- [x] \`[external] fetch\` has \`externalKind: "http"\`, edge has \`callType: "awaited"\`
- [x] \`analyze_codebase\` — \`topRefactorIssues\` contains no \`file: "external"\` entries
- [x] \`spec-gen setup --tools claude\` — \`.claude/settings.json\` gets the analyze hook
- [x] Touch a source file → \`analyze_codebase\` re-analyzes; leave files unchanged → instant cache hit

### DIFF
--- a/src/api/run.test.ts
+++ b/src/api/run.test.ts
@@ -102,7 +102,12 @@ vi.mock('../core/generator/adr-generator.js', () => ({
   }),
 }));
 
+vi.mock('../core/services/mcp-handlers/utils.js', () => ({
+  isCacheFresh: vi.fn(),
+}));
+
 import { access, stat, readFile } from 'node:fs/promises';
+import { isCacheFresh } from '../core/services/mcp-handlers/utils.js';
 import { detectProjectType, getProjectTypeName } from '../core/services/project-detector.js';
 import { getDefaultConfig, readSpecGenConfig, writeSpecGenConfig, specGenConfigExists, openspecDirExists, createOpenSpecStructure } from '../core/services/config-manager.js';
 import { gitignoreExists, isInGitignore, addToGitignore } from '../core/services/gitignore-manager.js';
@@ -129,6 +134,7 @@ const mockGitignoreExists = vi.mocked(gitignoreExists);
 const mockIsInGitignore = vi.mocked(isInGitignore);
 const mockAddToGitignore = vi.mocked(addToGitignore);
 const mockCreateLLMService = vi.mocked(createLLMService);
+const mockIsCacheFresh = vi.mocked(isCacheFresh);
 
 // ============================================================================
 // FIXTURES
@@ -185,6 +191,7 @@ function setupMocks({ configExists = false, analysisRecent = false } = {}) {
 
   // Analysis mocks
   const mtime = analysisRecent ? RECENT_MTIME : OLD_MTIME;
+  mockIsCacheFresh.mockResolvedValue(analysisRecent);
   mockAccess.mockResolvedValue(undefined);
   mockStat.mockResolvedValue({ mtime } as Awaited<ReturnType<typeof stat>>);
   mockReadFile.mockImplementation((path) => {
--- a/src/api/run.ts
+++ b/src/api/run.ts
@@ -7,9 +7,10 @@
  */
 
 import { join } from 'node:path';
-import { readFile, stat, mkdir, writeFile } from 'node:fs/promises';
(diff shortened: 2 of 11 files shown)


clay-good:
Thank you @laurentftech ❤️

Issue #52, opened 2026-04-03
Getting generate error "Invalid schema for response_format 'response'" with OpenAI


eldare:
```
spec-gen generate                                                                                  ✭ ✱

Loading Configuration

  Project: nodejs
  OpenSpec path: ./openspec


Loading Analysis

🔍 Using analysis from 10 minutes ago
  Files analyzed: 106
  Domains detected: server, client, plan, scripts, expansion, scraper, accessibility, services, api, tailwind-config-cjs, expansion-plan, original-plan, scrapers, public, blog, composables, components, src


Pre-flight Checks

  Model: gpt-4o
  Estimated tokens: 90,239
🧠 Estimated cost: ~$0.42

⚠ Existing specs will be replaced (backed up)


Generating Specifications

⚠ Stage 2: failed to analyze client/src/types/index.ts: {
  "error": {
    "message": "Invalid schema for response_format 'response': schema must be a JSON Schema of 'type: \"object\"', got 'type: \"array\"'.",
    "type": "invalid_request_error",
    "param": "response_format",
    "code": null
  }
}
```

PR #73, opened 2026-04-28
fix(llm): normalize OpenAI structured output schemas

Wars:
## Summary

- Normalize OpenAI/OpenAI-compatible `response_format.json_schema.schema` payloads before sending structured output requests.
- Close every object schema, require all object properties, and preserve optional semantics by allowing `null`.
- Keep `CopilotProvider` on the existing array-wrapper behavior.
- Add LLM spec coverage for the OpenAI structured output normalization contract.

## Problem

`spec-gen generate` could fail during the OpenAI generation flow when stage schemas were sent as structured outputs. In Stage 2, OpenAI returned HTTP 400 errors like:

```text
Invalid schema for response_format 'response': In context=(), 'additionalProperties' is required to be supplied and to be false.
```

## Implementation

This adds provider-local schema normalization for OpenAI structured outputs:

- Top-level array schemas are still wrapped as an object with an `items` property.
- Every object schema is recursively given `additionalProperties: false`.
- Every object schema gets `required` set to all keys under `properties`.
- Properties that were optional in the caller-provided schema are made nullable in the provider schema.
- The caller-provided schema is deep-cloned before normalization so prompts, validation, and non-OpenAI paths still use the original schema.
- `OpenAIProvider` and `OpenAICompatibleProvider` use the normalized schema.
- `CopilotProvider` is unchanged.

## Tests

- `npm run test:run -- src/core/services/llm-service.test.ts`
- `npm run typecheck`
- `openspec validate llm`
- `git diff --check`
- Manually verified built `spec-gen generate` with the OpenAI provider now completes successfully.

## Notes

Follow-up to #52: that change handled OpenAI's top-level array root restriction; this PR handles the remaining OpenAI Structured Outputs schema requirements for object schemas.

`openspec validate --all` still reports an unrelated existing `spec/analyzer` failure; `spec/llm` validates successfully.

### DIFF
--- a/openspec/specs/llm/spec.md
+++ b/openspec/specs/llm/spec.md
@@ -345,6 +345,42 @@ The system SHALL include `cursor-agent` in provider pricing lookup with zero per
 - **WHEN** `lookupPricing` is called for provider `cursor-agent` and any model identifier
 - **THEN** returned input and output per-million token rates are zero
 
+### Requirement: OpenAI structured output schema normalization
+
+The system SHALL normalize JSON schemas sent through OpenAI and OpenAI-compatible `response_format.json_schema.schema` so that every object schema is closed, every object property is required, and properties that were optional in the caller schema remain nullable in the provider schema.
+
+#### Scenario: OpenAI object schemas are closed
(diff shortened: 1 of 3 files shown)


clay-good:
Thank you @Wars ❤️

PR #81, opened 2026-05-15
fix(decisions): auto-clean legacy spec-gen hook on install/uninstall

laurentftech:
## Summary

- `installPreCommitHook` now strips any `spec-gen-decisions-hook` block it finds — even when the openlore block is already present (silent migration for existing repos)
- `uninstallPreCommitHook` also removes the legacy block alongside the openlore one
- Fixes double-gate-execution introduced by the abrupt spec-gen → openlore rename

## Test plan

- [ ] Clone a repo with a pre-commit hook containing both `spec-gen-decisions-hook` and `openlore-decisions-hook` blocks
- [ ] Run `openlore decisions --uninstall-hook` — verify both blocks removed
- [ ] Run `openlore setup --tools claude` on a repo with only the old `spec-gen-decisions-hook` — verify it's replaced with the openlore block only
- [ ] Run on a fresh repo — no regression


clay-good:
Thank you @laurentftech ❤️

PR #118, opened 2026-06-02
fix: move all soft-loaded grammars to optionalDependencies

laurentftech:
## Problem

All tree-sitter grammars loaded via `loadGrammarSoft()` can fail silently at runtime — `loadGrammarSoft` catches import errors and returns empty results with a warning when the module is absent. This means they are **optional by contract**.

However, 7 of them were declared in `dependencies` (hard install). On machines where `node-gyp` cannot compile native addons (corporate proxy, TLS inspection, offline CI), `npm install` fails entirely, blocking openlore adoption.

Previously discovered: `tree-sitter-kotlin` (PR from fork). Full audit reveals 6 more packages with the same issue.

## Fix

Move all `loadGrammarSoft`-loaded grammars to `optionalDependencies`:

| Package | Before | After |
|---------|--------|-------|
| `tree-sitter-bash` | `dependencies` | `optionalDependencies` |
| `tree-sitter-c` | `dependencies` | `optionalDependencies` |
| `tree-sitter-c-sharp` | `dependencies` | `optionalDependencies` |
| `tree-sitter-elixir` | `dependencies` | `optionalDependencies` |
| `tree-sitter-kotlin` | `dependencies` | `optionalDependencies` |
| `tree-sitter-php` | `dependencies` | `optionalDependencies` |
| `tree-sitter-scala` | `dependencies` | `optionalDependencies` |
| `tree-sitter-swift` | already optional | ✓ |

## Regression guard

Adds `grammar-optional-deps.test.ts`: statically parses `call-graph.ts` to extract every `loadGrammarSoft()` import target and asserts it is in `optionalDependencies`. Future additions of soft-loaded grammars will fail CI if not declared optional.

## Test plan

- [ ] `npm install openlore` succeeds in a restricted network environment
- [ ] All soft-loaded language analyses still work where native builds succeed
- [ ] Warning (not fatal error) displayed when a grammar is unavailable
- [ ] `grammar-optional-deps.test.ts` passes

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,18 +98,12 @@
     "react": "^19.1.1",
     "react-dom": "^19.1.1",
     "tree-sitter": "^0.22.4",
-    "tree-sitter-bash": "^0.23.3",
-    "tree-sitter-c": "^0.23.6",
-    "tree-sitter-c-sharp": "^0.23.1",
     "tree-sitter-cpp": "^0.23.4",
-    "tree-sitter-elixir": "^0.3.5",
     "tree-sitter-go": "^0.23.4",
     "tree-sitter-java": "^0.23.5",
-    "tree-sitter-php": "^0.23.12",
     "tree-sitter-python": "^0.23.6",
     "tree-sitter-ruby": "^0.23.1",
     "tree-sitter-rust": "^0.24.0",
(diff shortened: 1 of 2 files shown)


clay-good:
Thank you @laurentftech ❤️

PR #126, opened 2026-06-04
feat(serve): warm openlore daemon + Pi extension

laurentftech:
## Problem

When multiple clients (Claude Code MCP, Cline, Pi) connect to the same repo, each spawns its own watcher and its own caches. Two agents race to write `.openlore/analysis`. Call-graph analysis drifts silently after edits because watchers only refresh signatures and vectors, never the full graph.

HTTP is a transport choice, not a second API surface. The goal is shared process ownership across heterogeneous clients — one warm process, one watcher, one source of truth per repository.

## Solution

`openlore serve` is a long-lived loopback HTTP daemon. All transports discover and delegate to it, so there is exactly one warm process per repository, shared across agents.

```
Pi extension   ─┐
Claude Code MCP ─┤──► openlore serve ──► dispatchTool() ──► handlers
Cline MCP      ─┘         │
                     one watcher
                     one cache
```

## Changes

**`tool-dispatch.ts` (new)** — single `dispatchTool()` called by both the stdio MCP server and the HTTP daemon. The two transports share one dispatch table and cannot drift. No behavior change; existing MCP tests pass unchanged.

**`openlore serve`** — loopback HTTP daemon:
- `GET /health` → `{ ok, version, preset, tools, uptimeMs }`
- `POST /tool/:name` body `{ directory, args }` → handler result
- Writes `.openlore/serve.json` for discovery; `--stop` for clean shutdown
- Optional `--token` (`x-openlore-token` header)
- Refuses to start a second daemon for a root already served by a live one

**Watcher + re-analyze lane** — `McpWatcher` keeps signatures/vectors live; a new `onBatchFlushed` hook schedules a debounced `analyze --force` after each edit burst, closing the call-graph drift gap between commits.

**MCP delegation** (`serve-client.ts`) — on each tool call, MCP discovers a running daemon and forwards to it. Falls back to in-process dispatch transparently on transport failure. Default: discover-and-reuse only (no spawn, no background process left behind). `--daemon` flag opts in to spawning.

**Watcher invariant enforced in both MCP paths** — both the `--watch-auto` (auto-detect) and `--watch <dir>` (explicit) paths skip watcher creation when a daemon is already serving the directory.

**Pi extension** (`examples/pi/openlore.ts`) — auto-starts the daemon, injects architecture digest + spec index + task-grounded `orient` on the first turn, registers 7 navigation tools. Uses `fetch` + `AbortSignal`. No MCP, no subprocess per call.

**`openlore setup --tools pi [--global]`** — installs the extension to `.pi/extensions/` or `~/.pi/agent/extensions/`.

**Auto-heal on schema bump** — if EdgeStore resets after a version upgrade, the daemon waits for background rebuild before serving graph tools instead of returning "index is empty".

## Tests

| File | What it covers |
|------|----------------|
| `serve.test.ts` | health, preset advisory, `serve.json` lifecycle, token gate, dup-daemon guard, stale `--stop`, 20 concurrent calls, watcher-invariant reuse path |
| `serve-client.test.ts` | discover / call / unknown-tool / unreachable / kill-9 simulation (stale serve.json → null) |
| `mcp-watcher.test.ts` | `onBatchFlushed` fires on real change, not on no-op |

Full suite: 150 test files, 3164 tests, 0 failures.

## Verified end-to-end

- orient, search_code, subgraph all return correct results from a warm daemon
- 20 concurrent calls: 20/20 valid responses, daemon healthy after
- Branch switching (main ↔ feat branch): daemon survives, orient coherent post-switch
- Kill -9: stale serve.json left on disk; health probe fails cleanly; new daemon starts and overwrites serve.json with updated PID
- Multi-client: MCP `--daemon` + running daemon → 1 serve.json, orient succeeds via delegation

## Known limitations (out of scope)

- Re-analyze runs in-process (brief event-loop block). Moving it off-thread requires explicit cache invalidation after rebuild — the two are coupled and will ship together.
- Watcher listens to `change` not `unlink`; stale rows for deleted files linger until the next `analyze --embed`. Pre-existing limitation.

### DIFF
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -213,7 +213,7 @@ openlore serve --stop                   # stop the daemon serving this directory
 | `-d, --directory <path>` | Project root to serve; discovery file written here (default: cwd) |
 | `-p, --port <number>` | Port to bind (default: ephemeral free port) |
 | `--host <host>` | Host to bind (default: `127.0.0.1`) |
-| `--preset <name>` | Tool surface: `minimal`, `navigation` (default), or `all` |
+| `--preset <name>` | Advisory surface reported by `/health`: `minimal`, `navigation` (default), or `all`. The daemon dispatches any known tool regardless; clients curate their own surface |
 | `--token <token>` | Require this token as the `x-openlore-token` header (default: `$OPENLORE_SERVE_TOKEN`) |
 | `--no-watch` | Disable the freshness watcher + re-analyze lane |
 | `--stop` | Stop a running daemon for `--directory` and exit |
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -18,8 +18,9 @@
 | `openlore decisions --install-hook` | Install the pre-commit hook that gates commits until decisions are reviewed | No |
 | `openlore run` | Full pipeline: init, analyze, generate | Yes |
 | `openlore view` | Launch interactive graph & spec viewer in the browser | No |
-| `openlore setup` | Install workflow skills into the project (Vibe, Cline, GSD, BMAD) | No |
+| `openlore setup` | Install workflow skills into the project (Vibe, Cline, GSD, BMAD, Pi) | No |
 | `openlore mcp` | Start MCP server (stdio, for Cline / Claude Code) | No |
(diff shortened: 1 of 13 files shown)


clay-good:
Thank you @laurentftech ❤️

PR #133, opened 2026-06-07
chore: fix deprecated Language.query() and bump glob to v13

laurentftech:
## Summary

- Replace deprecated `Language.query(src)` with `new Query(language, source)` in native tree-sitter path (`safeQuery` already used the new API; fixes the remaining call site)
- Use `web-tree-sitter`'s own `Query` constructor in the WASM path (Dart/Lua grammars) — previously `lang.query()` triggered the same deprecation warning; switching to `new WasmQuery(lang, src)` silences it without breaking WASM grammar isolation
- Bump `glob` from `^11.0.0` to `^13.0.6` (latest) — silences npm deprecation warning on install
- Update package `description` to reflect current positioning — visible as subtitle on https://pi.dev/packages/openlore

## Test plan

- [ ] `tsc --noEmit` passes (0 errors)
- [ ] `eslint src` passes (0 errors)
- [ ] `vitest run` passes — including `wasm-multigrammar.test.ts` (Dart + Lua graphing)
- [ ] `npm install` produces no deprecation warnings for glob or tree-sitter


clay-good:
Thank you @laurentftech ❤️

PR #134, opened 2026-06-08
hotfix(pi): use shell:true on Windows to avoid ENOENT on spawn

laurentftech:
## Summary

- `ensureDaemon`: add `shell: true` on `win32` — npm installs CLI wrappers as `.cmd` files on Windows, which `spawn()` can't resolve without shell delegation
- `trySpawn` / `runConfigWizard`: same `shell: true` on `win32` so both `openlore analyze` and `npx openlore analyze` resolve correctly
- Bump `@earendil-works/pi-coding-agent` + `@earendil-works/pi-ai` to `0.79.0`
- Use `getAgentDir()` from the new 0.79 SDK exports for the global install path — respects env var overrides instead of hardcoding `~/.pi/agent`

## Context

Without this fix, launching Pi on Windows throws ENOENT on spawn — the extension fails to start the openlore daemon entirely.

**0.79 bonus**: the config wizard (`openlore_configure`) uses `ui.select` / `ui.input` / `ui.confirm` throughout. In 0.78.x those dialog calls silently no-oped in RPC mode. With 0.79's extension UI protocol they emit `extension_ui_request` on stdout and block for a response — the wizard now works interactively in IDE / headless contexts, not just TUI. No code change needed; the bump enables it automatically.

## Test plan

- [ ] On Windows: `pi install npm:openlore` → daemon starts without ENOENT
- [ ] On Windows: config wizard runs `openlore analyze` successfully
- [ ] On macOS/Linux: no regression — platform checks are `win32`-gated
- [ ] In RPC mode (IDE): `openlore_configure` tool shows interactive dialogs

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -41,8 +41,8 @@
         "openlore": "dist/cli/index.js"
       },
       "devDependencies": {
-        "@earendil-works/pi-ai": "^0.78.1",
-        "@earendil-works/pi-coding-agent": "^0.78.1",
+        "@earendil-works/pi-ai": "^0.79.0",
+        "@earendil-works/pi-coding-agent": "^0.79.0",
         "@eslint/js": "^9.17.0",
         "@types/node": "^22.10.5",
         "@vitest/coverage-v8": "^4.0.18",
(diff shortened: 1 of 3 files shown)


clay-good:
Thank you @laurentftech ❤️

PR #136, opened 2026-06-09
fix(serve) + feat(pi): Windows console fix + config wizard UX refactor

laurentftech:
### fix(serve): hide daemon console window on Windows
- Add `windowsHide: true` to `spawn()` in `serve-client.ts`
- Prevents visible terminal windows when the serve daemon spawns on Windows
- No effect on macOS/Linux

### feat(pi): refactor config wizard to flat non-sequential menu
- Replace `configureGeneration` + `configureEmbedding` sub-wizards with a single flat loop
- Every field is independently editable — no forced sequential traversal
- Fix silent bug: ESC on a sub-select previously overwrote the value with the list default
- Add section separators (`── Generation (LLM)`, `── Embedding (retrieval)`, `── Analysis`)
- Add `Discard & close` as explicit cancel action (was previously implicit ESC-only)
- Add `Remove embedding` entry when embedding is configured
- Condition "Run openlore analyze now?" only when provider or model actually changed
- Unify field label format with `padEnd(14)` across all entries

## Test plan

- [ ] On Windows: no console window appears when daemon spawns
- [ ] Config wizard: changing only model does not require re-selecting provider
- [ ] Config wizard: ESC on a sub-select returns to menu without modifying the field
- [ ] Config wizard: "Run analyze now?" only appears after provider/model change
- [ ] Config wizard: "Remove embedding" removes embedding config
- [ ] Config wizard: "Discard & close" exits without saving
- [ ] macOS/Linux: no regression

### DIFF
--- a/src/core/services/serve-client.ts
+++ b/src/core/services/serve-client.ts
@@ -98,7 +98,7 @@ export async function ensureServeDaemon(
     const child = spawn(
       process.execPath,
       [cli, ...serveSpawnArgs(directory)],
-      { cwd: directory, stdio: 'ignore', detached: true },
+      { cwd: directory, stdio: 'ignore', detached: true, windowsHide: true },
     );
     child.on('error', () => {}); // swallow — caller falls back to in-process
     child.unref();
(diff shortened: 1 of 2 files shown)


clay-good:
Thank you @laurentftech ❤️

PR #142, opened 2026-06-11
feat(call-graph): synthesized dynamic-dispatch edges — multi-language event/callback/actor synthesis with provenance

clay-good:
Implements `openspec/changes/add-synthesized-dynamic-dispatch-edges/`: closes the call graph's dynamic-dispatch blind spots **deterministically** (no LLM, AST pattern-matching only), with every recovered edge **labeled by its provenance** so consumers see exactly which conclusions rest on a heuristic vs direct name resolution.

## Why
The call graph is built purely by direct name resolution, so it's blind to callbacks, events, framework routes, and message dispatch. The cost is concrete: `find_dead_code` reported live handlers as **high-confidence dead**, `analyze_impact` under-counted blast radius, and `select_tests` omitted tests reaching changed code only through a handler — the exact false positives `reachability.ts`'s own header documented.

### Edge provenance (foundation)
- `EdgeConfidence` gains `'synthesized'`; `CallEdge` gains `synthesizedBy` (the rule name); `CALL_DISTANCE_COSTS`/`callDistance` give synthesized edges a strictly-higher cost than any directly-resolved confidence, so `find_path` prefers a resolved route.
- The SQLite edge store persists `synthesized_by` (schema 5→6, rebuild-on-open) so provenance survives into the edge-store-backed tools; `get_subgraph` flags synthesized edges.
- **Direct edges are byte-for-byte unchanged**; graphs without `synthesizedBy` load unchanged.

### Five synthesis rules (additive post-resolution pass; shared pairing/fan-out/provenance core)
| Rule (`synthesizedBy`) | Pattern | Languages |
|---|---|---|
| `event-channel` | `on/emit`-style string/symbol/constant key | JS/TS, Python, Ruby, PHP, Swift (NotificationCenter) |
| `type-event` | annotated/typed handler ↔ `new T()` dispatch | Java, C#, Kotlin |
| `route-handler` | route → its bound handler | via existing route inventory (Python/TS/Java) |
| `callback-registration` | named handler → curated registrar the runtime invokes | Go (`net/http` + gin/echo/chi), JS/TS schedulers, C++ (Qt `connect` slots) |
| `actor-message` | message dispatch ↔ handler by message tag | Elixir GenServer (`cast`/`call`/`send` ↔ `handle_*`) |

Handler shapes covered per language include bare/member/`self.`/`this.` refs, `.bind()`, blocks, lambdas/closures (wired to the functions their body calls), PHP array callables, C++ pointer-to-member slots, and constant/template/symbol keys (namespaced so a string never pairs with a same-text symbol/constant). Pairing is **per-language** (no cross-language pairing); per-channel fan-out is **capped** (over-cap channels dropped, not guessed).

### Provenance-aware reachability + strict mode
- Synthesized edges are **traversed by default**, so callback/event/route/actor handlers are no longer false dead-positives (retires the documented limit). A candidate reachable only via a synthesized edge whose dispatcher is itself dead is downgraded to `low` with the rule named.
- A **`directResolvedOnly` strict mode** is threaded through `find_dead_code`, `analyze_impact`, `get_subgraph`, `find_path`, `select_tests`, and `trace_execution_path` (trade completeness for certainty).
- **Structural-metric isolation:** synthesis augments *reachability* only — synthesized edges are excluded from `fanIn`/`fanOut` (hence hub/god/entry classification), from `get_health_map` betweenness + untested-degree, and from `get_surprising_connections`, so one heuristic edge can't manufacture a false hub/bridge/surprise.

### Precision / correctness fixes
- **Runtime callback locals are never resolved as handlers.** Reference-based handler resolution skips `resolve`/`reject`/`next`/`done`/`callback`/`cb`/`err`/`error`/`fulfill` — adversarial review on the real OpenLore repo (~3,900 functions) found these were the rules' *only* false positives (`setTimeout(resolve, ms)` inside `new Promise((resolve) => …)` wiring to a coincidentally same-named function). One chokepoint in the shared `resolveHandler` closure; synthesized edges on the repo go from 2 (both false) to 0. Regression-tested.
- Dead-code liveness roots now recognize C#'s capitalized `Main` (was only `main`/`default`).

## Deliberately NOT covered (the no-guessing discipline)
Mechanisms that expose no statically-pairable key+handler / type+handler / tag+handler — **Go channels, Akka/Scala `receive` actors, Rust, C, Scala, Lua, Bash, Dart** (no Dart grammar) — get **no collector**. The pass emits nothing for them rather than fabricate an edge. Reflection and computed/string-built dispatch (`obj[name]()`) likewise remain a documented blind spot.

## Verification
- **Full `src` + `examples` suites green: 3478 passed, 2 skipped**; `tsc --noEmit` clean; `eslint src` clean; clean `dist` build OK.
- `edge-synthesis.test.ts` (62 cases: every rule + negatives — mismatched key/type/tag, non-registrar, inline-closure, un-annotated, namespace isolation, per-language isolation, C++ signal-not-wired, runtime-local-not-wired regression), `reachability-synthesis.test.ts`, `edge-store.test.ts` (provenance round-trip), `call-graph.test.ts` (distance exhaustiveness).
- **Adversarial run on the real OpenLore repo** (analyze + inspect every synthesized edge): 0 false positives after the fix.
- **Comprehensive mixed-language e2e through the compiled CLI**: all rules fire on one analyzed repo and every core tool (`orient`, `find_dead_code` incl. strict, `analyze_impact`, `get_subgraph`, `get_call_graph`, `get_health_map`, `get_surprising_connections`, `get_critical_hubs`, `find_path`, `trace_execution_path`, `select_tests`) runs cleanly with the direct call graph intact.
- No new MCP tool; default/`minimal` tool surface unchanged in size. All changes confined to `src/` + the change dir; `main` untouched.

## Notes
- The decisions gate was skipped (`--no-verify`) at the author's direction; consolidated decisions remain recorded locally (gitignored), not synced to specs.
- Pre-existing, unrelated: a NUL byte in `orient.ts` (flagged separately) — untouched here.


clay-good:
## Deep adversarial analysis + fix (commit 46f24bd)

Rather than re-run the green suite, I stress-tested the synthesis against the **real OpenLore codebase** (~3,900 functions) — analyze, then inspect *every* synthesized edge for false positives. That surfaced one real bug:

**Found:** `sleep → resolve` and `loadAttempt → resolve` (callback-registration). Root cause: `setTimeout(resolve, ms)` inside `new Promise((resolve) => …)` — `resolve` is the executor's **parameter**, but reference-based handler resolution matched it (via the global-uniqueness fallback) to an unrelated arrow `const resolve = …` in a *different* file. A genuine false positive.

**Fixed:** the shared `resolveHandler` closure — the single chokepoint every reference-based rule (event-channel, callback-registration, route-handler) flows through — now skips runtime/promise/middleware callback **locals** (`resolve`, `reject`, `next`, `done`, `callback`, `cb`, `err`, `error`, `fulfill`). These are never registered named handlers, so the fix removes false edges without dropping real ones. (`type-event`/`actor-message` resolve handlers structurally, not by reference, so they were never affected.)

**Result:** synthesized edges on the OpenLore repo went **2 (both false) → 0** — correct, since the repo has no genuine event/callback-with-named-handler patterns. Regression-tested both ways (Promise-`resolve` local not wired; a genuinely-named scheduler handler still wired).

### Final verification
- Full `src` + `examples` suites: **3478 passed, 2 skipped, 0 failed**; `tsc` clean; `eslint src` clean; clean `dist` build OK.
- Adversarial real-repo run: **0 false positives**.
- Mixed-language e2e (all 5 rules + every core tool) clean; direct call graph intact.
- Working tree clean; branch synced; all changes in `src/` + the change dir; `main` untouched.

PR title + description now reflect the complete state (15 commits): five rules across JS/TS, Python, Ruby, PHP, Swift, Java, C#, Kotlin, Go, C++, Elixir; full provenance; strict mode across six traversal tools; structural-metric isolation; the runtime-locals precision fix; and the C# `Main` fix — with the deliberate no-collector boundary for unpairable mechanisms.

PR #145, opened 2026-06-12
chore(homebrew): remove non-native prebuilds after install

rzylim:
tree-sitter adds prebuild binaries for multiple cpu architectures which is not allowed under homebrew's single-architecture policy.

This pull request contains my naive fix which modifies the formula to remove all but native prebuilds in `openlore_installation_directory/lib/node_modules/openlore/**/prebuilds/`. Please let me know if any changes need to be made!

### DIFF
--- a/packaging/homebrew/openlore.rb
+++ b/packaging/homebrew/openlore.rb
@@ -20,6 +20,11 @@ class Openlore < Formula
 
   def install
     system "npm", "install", *std_npm_args
+    os = OS.mac? ? "darwin" : "linux"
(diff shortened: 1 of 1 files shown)


rzylim:
Also, would you mind if I submitted Openlore for homebrew/core consideration on your behalf? The threshold for notability is lower at 75 stars if it's not a self-submit, instead of 225 stars for a self-submit.

clay-good:
Hi @rzylim, Thank you! Yes, please. That's very kind of you. That was my main concern, I thought we needed to be a larger repo before they'd allow it.

clay-good:
Thank you @rzylim ❤️

PR #147, opened 2026-06-13
feat: make native tree-sitter deps optional (silent install without build tools)

laurentftech:
## Problem

`pi install npm:openlore` runs `npm install openlore` at every Pi coding agent startup. On environments without native build tools (corporate Windows, Nexus proxy), this produces:
- `npm warn allow-scripts` noise for 17+ packages
- `code: 1` failures for `tree-sitter-cli` and `tree-sitter-kotlin`
- ~5s overhead on every agent startup

## Solution

Move `tree-sitter` and all language grammar packages from `dependencies` to `optionalDependencies`. npm tolerates optional dep failures silently — no more alarming output for users who only need the Pi extension (which has zero native imports itself).

### Changes

**`package.json`** — move tree-sitter + all language grammars to `optionalDependencies` (alongside existing `@lancedb/lancedb`). Pure JS/WASM alternatives (`web-tree-sitter`, `tree-sitter-wasms`) stay in `dependencies`.

**`src/core/analyzer/ast-chunker.ts`** and **`src/core/analyzer/call-graph.ts`** — replace static `import Parser from 'tree-sitter'` with lazy dynamic imports using the `undefined/null` sentinel pattern:
- `undefined` = not yet tried
- `null` = tried, unavailable (falls back gracefully)
- All 12 parser getter functions and all extraction paths are null-safe

When native tree-sitter is absent, AST chunker falls back to blank-line chunking; call-graph extraction returns empty results for affected files.

## Test plan

- [x] `npm run typecheck` — 0 errors
- [x] `npm run test:run` — 3536 tests pass
- [ ] Install in an environment without build tools: `npm install openlore` completes without `code: 1` failures


clay-good:
Thank you @laurentftech ❤️

PR #157, opened 2026-06-17
fix(serve): idle self-shutdown + keepalive so orphaned daemons can't pile up in RAM

laurentftech:
## Problem

On Windows, opening and closing Pi repeatedly leaves ghost `openlore serve` processes that accumulate and exhaust RAM.

**Mechanism:** the warm daemon is spawned **detached** by clients (Pi extension, MCP server) and is not a child of any of them — closing a client never signals it. A slow/flaky first `/health` probe (cold Node HTTP server on Windows often exceeds the 1 s timeout) makes a client *and* the existing single-instance guard misread a live daemon as dead. They spawn a fresh daemon, overwrite `serve.json`, and **orphan** the previous one. Orphans hold their port + caches forever and are never referenced again → pileup.

## Fix

1. **Idle self-shutdown** (`serve`): a timer tears down and exits after N minutes with no requests (default **15**, `--idle-timeout <minutes>`, `0` disables). Any request — including `/health` — resets it. Orphans get zero requests and reap themselves; the in-use daemon is kept alive by tool calls. Timer is `unref`'d and cleared on teardown.
2. **Keepalive** (Pi extension): while a session is open, ping the known daemon's `/health` every **5 min** (well under the idle window) so the in-use daemon never idle-reaps mid-session. Cleared on `session_shutdown`; on a hard kill the interval dies with the host process — either way pings stop and the now-unused daemon reaps. **Only daemons the extension knows are pinged**, so orphans still get zero pings and still reap — keepalive cannot resurrect the pileup it's paired with.
3. **Bump the `/health` probe timeout** 1000 ms → 2500 ms at **all three call sites** (serve single-instance guard, Pi extension, MCP `serve-client`) so a cold server on Windows isn't misread as dead — fewer orphans created.

**Invariant:** the idle timeout stays ~3× the keepalive interval (15 min vs 5 min), so two consecutive missed pings can't wrongly reap an in-use daemon. Lowering one without the other is called out in the code comments.

Cross-platform: idle shutdown bounds every daemon's lifetime regardless of client behaviour or OS signal quirks (on Windows `process.kill`/SIGTERM on a detached, shell-wrapped PID is unreliable). The fix deliberately does **not** rely on signals.

## Lifecycle summary

```
Pi open    → daemon warm, keepalive pings every 5 min → never idle-reaps
Pi closes  → session_shutdown clears keepalive (or process dies) → pings stop
15 min later (no tool calls, no pings) → daemon idle-reaps, serve.json removed
Pi reopens → serve.json gone/dead → one clean spawn
orphan (never in serve.json) → never pinged, never called → reaps on its own
```

## Scope: why keepalive is Pi-only (not the MCP server)

Keepalive lives with whoever **owns** (spawned) the daemon and wants it warm across idle gaps. That's Pi.

The MCP server (KiloCode, Claude Code, Cline) behaves differently:

- **By default it does not spawn a daemon at all** — it reuses one only if already running (e.g. started by Pi or an explicit `openlore serve`), otherwise it dispatches **in-process**. So a plain MCP session creates no detached process and contributes nothing to the pileup.
- When it reuses Pi's daemon, that daemon is **already kept warm by Pi's keepalive**.
- During active use, every tool call (`POST /tool/*`) resets the idle timer, so the daemon stays warm without any keepalive.

The only gap an MCP-side keepalive would close is `openlore mcp --daemon` running solo (no Pi) and sitting idle past the timeout — which costs a single respawn on the next call, not a pileup. Adding an interval + lifecycle teardown to `serve-client`/`mcp.ts` for that niche isn't worth the surface. Correctness is unaffected either way: idle-reap + respawn always works; keepalive is purely a warm-start optimisation for the interactive Pi case. The `/health` timeout bump (item 3) **is** applied to the MCP path, since that prevents orphan *creation* and is cheap.

## Changes

- `src/cli/commands/serve.ts` — idle timer + `touchActivity()` reset per request; `--idle-timeout`; `idleTimeoutMs()` parser; `HEALTH_PROBE_TIMEOUT_MS`; default 15 min
- `src/pi/extension.ts` — keepalive interval (5 min, cleared on `session_shutdown`); health probe 1000 → 2500 ms
- `src/core/services/serve-client.ts` — health probe 1000 → 2500 ms (MCP path)
- `src/cli/commands/serve.test.ts` — `idleTimeoutMs` unit tests + functional idle tests (reap / timer-reset / disabled), `process.exit` stubbed so teardown runs without killing the runner

## Test

- 24/24 serve tests + 7/7 serve-client tests pass; `tsc --noEmit` + eslint clean; CI green
- Functional idle tests assert: serve.json removed after timeout, daemon survives while pinged, never reaps when disabled

### DIFF
--- a/src/cli/commands/serve.test.ts
+++ b/src/cli/commands/serve.test.ts
@@ -7,13 +7,13 @@
  * .openlore. We assert transport behaviour, not handler output.
  */
 
-import { describe, it, expect, afterEach } from 'vitest';
+import { describe, it, expect, afterEach, vi } from 'vitest';
 import { mkdtemp, rm, readFile, access, mkdir, writeFile } from 'node:fs/promises';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 import { request as httpRequest } from 'node:http';
 import { DatabaseSync } from 'node:sqlite';
-import { startServe, readDescriptor, type ServeHandle } from './serve.js';
+import { startServe, readDescriptor, idleTimeoutMs, type ServeHandle } from './serve.js';
 import { EdgeStore } from '../../core/services/edge-store.js';
 import { openloreAnalyze } from '../../api/analyze.js';
 import { OPENLORE_DIR, OPENLORE_ANALYSIS_SUBDIR } from '../../constants.js';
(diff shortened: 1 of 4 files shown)


clay-good:
Thank you @laurentftech 🙏

PR #159, opened 2026-06-17
fix(select_tests): default to working-tree diff vs HEAD when no args given

laurentftech:
## Problem

`select_tests` required either `changedSymbols` or `diffRef`; with neither it returned `{ error: "Provide changedSymbols or diffRef." }`. Weak tool-callers (a local model in Pi, but also any agent) frequently call it with **no arguments** — and got an opaque error with no path forward. Observed with codestral in the Pi extension: it invoked `select_tests` bare and the result was incomprehensible.

## Fix

When neither `changedSymbols` nor `diffRef` is given, **default to diffing the working tree against HEAD** — the tool's most common intent ("which tests cover my current changes?"). A bare call now does the useful thing instead of erroring, and the result is self-explanatory:

```jsonc
// select_tests({ directory })  — no other args
{
  "changed": ["src/foo.ts"],
  "selectedTests": [ /* tests reaching those changes */ ],
  "note": "Called without changedSymbols/diffRef — diffed the working tree against HEAD. Pass changedSymbols or diffRef to target a specific change."
}
```

When the tree is clean the empty result's `message` also states it defaulted, so it's never mysterious.

This is daemon-side, so every client benefits — including the Pi extension (which forwards to the daemon) without any extra change.

## Changes

- `src/core/services/mcp-handlers/test-impact.ts` — precedence `changedSymbols → diffRef → HEAD`; `note` + empty-diff `message` flag the default
- `src/cli/commands/mcp.ts` — `changedSymbols`/`diffRef` documented as optional with the HEAD default
- `src/core/services/mcp-handlers/test-impact.test.ts` — replace the "errors with no args" test with: bare call defaults to HEAD (asserts `getChangedFiles` called with `baseRef: 'HEAD'` + `note`), and empty-diff message flags the default

## Test

- 9/9 test-impact tests pass; `tsc --noEmit` clean
- Verified end-to-end through the real `mcp__openlore__select_tests` tool against a daemon on this build: a no-arg call defaults to HEAD with the explanatory note, instead of the old error

### DIFF
--- a/src/cli/commands/mcp.ts
+++ b/src/cli/commands/mcp.ts
@@ -504,11 +504,11 @@ export const TOOL_DEFINITIONS = [
         changedSymbols: {
           type: 'array',
           items: { type: 'string' },
-          description: 'Changed function/method names. Provide this OR diffRef.',
+          description: 'Changed function/method names. Optional — if neither this nor diffRef is given, defaults to your current working-tree changes vs HEAD.',
         },
         diffRef: {
           type: 'string',
-          description: 'Git ref to diff the working tree against (e.g. "HEAD", "main"). Provide this OR changedSymbols.',
+          description: 'Git ref to diff the working tree against (e.g. "HEAD", "main"). Optional — defaults to HEAD when neither this nor changedSymbols is given.',
         },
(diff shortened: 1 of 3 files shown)


clay-good:
Thank you @laurentftech ❤️

PR #166, opened 2026-06-19
feat(verify-claim): structural claim verification — the agent's claims about code become auditable

clay-good:
## What

Adds the **`verify_claim`** MCP tool (change: `add-structural-claim-verification`). It reverses the usual direction of OpenLore: instead of the server handing facts to the agent, the **agent submits a structured claim about code and gets back a deterministic verdict plus a citation receipt** it can show a human. This converts the agent's confident-wrong failure mode ("this function is dead", "Y calls Z", "this change is safe") into *checked-or-flagged*, and makes the agent's output auditable.

```
claim   = { kind: 'calls' | 'reaches' | 'dead' | 'impacts' | 'safe-to-change', subject, object? }
verdict = { verdict: 'confirmed' | 'refuted' | 'unverifiable', reason, receipt?, confidenceBoundary }
```

## How each kind is decided (graph computation, never an LLM — north star `c6d1ad07`)

| kind | computation |
|------|-------------|
| `calls` | a direct caller→callee edge exists (synthesized-only edge → confirmed but the dispatch boundary is flagged) |
| `reaches` | forward reachability subject ⇒ object, with the path in the receipt |
| `impacts` | backward reachability: object transitively calls subject, so changing subject can require changes in object |
| `dead` | mark-and-sweep reachability (reuses `deadCodeIds`), run twice — synthesized-inclusive vs directly-resolved — to separate truly-dead from reached-only-via-heuristic-dispatch |
| `safe-to-change` | no internal caller depends on subject (empty directly-resolved blast radius) |

## Receipt as citation

A `confirmed`/`refuted` verdict carries a receipt reusing the **grounding-certificate shape** (`AnchorContext.certificateForAnchor`) — subject/object symbol, file, line span, content hash — plus the **index commit**. The receipt's `contentHash` equals an independent hash of the cited span, so a human can audit the claim against the index at a named commit (test-guarded).

## `unverifiable` is first-class

When a claim's answer rests on a dispatch blind spot (a symbol reached only through synthesized dynamic-dispatch edges, or an unresolved/ambiguous symbol), the verdict is `unverifiable` with the boundary **named** — reusing `add-confidence-boundary-disclosure`. OpenLore never fabricates a decisive verdict to look authoritative.

## Surface

- Conclusion-shaped (verdict + bounded receipt, never a graph). `tool-contract.ts` = `conclusion`.
- Opt-in **`verify` preset** (`orient` + `search_code` + `verify_claim`); never in `minimal`, the first-run default, or `navigation`.
- tools/list payload ceiling bumped 55_000 → 56_500 (one new tool, ~900 B; documented in `mcp-presets.test.ts`).

## Tests & dogfooding

- 19 unit tests: every verdict (confirmed/refuted/unverifiable) for all 5 kinds, input validation, ambiguous/unresolved symbols, and a **real-edge-store test that the receipt hash matches an independent span hash** (task 2).
- Full suite green: **3955 passed, 2 skipped, 0 failures**; typecheck + lint clean.
- **Dogfooded against this repo's live index**: `calls`/`reaches`/`impacts` confirmed with real `via` paths and receipts (real content hashes, line spans, index commit `47be10f`); `calls`/`dead`/`safe-to-change` refuted with caller counts; `unverifiable` for an unknown symbol. `confidenceBoundary.complete=false` correctly reflected working-tree staleness vs the build commit.

## Note on the base branch

Stacked on **`feat/confidence-boundary-disclosure` (#165)** because the `unverifiable` verdict reuses that PR's `assembleBoundary`/`KnownUnknowableCrossing` primitives, which are not yet on `main`. Base will switch to `main` once #165 merges. Never touches `main`.

Decision: `9a8084b6`. Closes the `add-structural-claim-verification` proposal (marked BUILT).

---

## Follow-up: dogfooding caught a misaligned-receipt bug (commit `61a8e5f`)

Re-dogfooding `verify_claim` against this repo's live index (re-analyzed at `59713a1`) surfaced a real defect in the **receipt** — the feature's headline "auditable citation." The receipt for `handleVerifyClaim` cited **lines 335–431** (hashing from mid-doc-comment to mid-body) instead of the true **338–441**.

**Root cause (pre-existing, in shared `anchor-adapter.ts`):** tree-sitter's native binding reports node `startIndex`/`endIndex` as **UTF-16 code-unit** indices, not byte offsets. The grounding-certificate and freshness hashing sliced a `Buffer` by those indices, so in any file with multibyte characters before a span (this repo is full of `—`/`⇒`/`…`), the cited `contentHash` and `lineSpan` drifted. The original unit test passed only because baseline and current hash sliced the **same wrong bytes** — internally consistent, absolutely wrong.

**Fix:** a shared `sliceNodeSpan(content, node)` slices the JS string by code units, reused by both `certificateForAnchor` and the freshness view (preserving the receipt-hash == freshness-hash invariant). ASCII spans are byte-identical → only multibyte spans change (one-time re-baseline). The multibyte adversarial test now derives offsets from the **real `CallGraphBuilder` parse** (ground truth on the offset unit) and asserts the certificate cites the exact line range + a hash of the real function source.

**Re-verified end to end:** receipt now cites `338–441`; its `contentHash` (`9d490063017e6c36`) matches an independent hash of those exact lines. Full suite **3957 passed**, 2 skipped; lint + typecheck clean. Decision `55e7f8e4`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -22,6 +22,7 @@
 | Recording an architectural decision before writing code | `record_decision` |
 | Persisting a durable, code-anchored fact for later sessions | `remember` (opt-in `memory` preset) — anchors a note to a symbol/file so it self-invalidates |
 | Recalling what's known about code you're touching | `recall` (opt-in `memory` preset) — returns memories with a freshness verdict; never serves orphaned ones as authoritative |
+| About to assert a structural fact to a user ("X is dead", "Y calls Z", "this is safe to change") | `verify_claim` (opt-in `verify` preset) — verify the claim against the graph, then cite the receipt to the human; an `unverifiable` verdict means hedge or read the source |
 
 For all other cases (reading a file, grepping, listing files) use native tools directly.
 
--- a/openspec/changes/add-structural-claim-verification/proposal.md
+++ b/openspec/changes/add-structural-claim-verification/proposal.md
@@ -1,0 +1,76 @@
+# Structural claim verification: make the agent's claims about code auditable
+
+> Status: BUILT (2026-06-19) — `verify_claim` MCP tool shipped behind the opt-in `verify` preset.
+> See `src/core/services/mcp-handlers/claim-verification.ts` (+ `.test.ts`). Decision: `9a8084b6`.
+> Stacked on `feat/confidence-boundary-disclosure` (PR #165) for the confidence-boundary primitives.
+> Phase-2 set (built after the memory + dispatch changes). Consumes the `unverifiable` verdict
+> from `add-confidence-boundary-disclosure` and the grounding certificate from
+> `add-trust-calibrated-context-economy`.
+
+## Why
+
+`add-trust-calibrated-context-economy` gives the agent evidence in one direction: OpenLore → agent
+("here is proof this fact is current"). The reverse direction is missing and is just as valuable:
+agent → OpenLore. When an agent is about to tell a user "this function is dead," "Y calls Z," or "this
+change is safe," it is often pattern-matching, and it is sometimes wrong. There is no cheap way for the
+agent to **check a structural claim against the deterministic graph and get back a verdict plus a
+citation it can show the user.**
+
+This is the difference between an agent that asserts and an agent that *verifies, then cites*. It
+converts the agent's confident-wrong failure mode into checked-or-flagged, and it makes the agent's
+output auditable by the human — OpenLore becomes the citation layer for claims about code, not just a
+private context source.
+
+## What changes
+
+1. **A claim-verification capability.** The agent submits a structured structural claim —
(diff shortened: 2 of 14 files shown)


clay-good:
Landed in `main` as part of the integrated umbrella merge of #162–#167 (main commit d34388e; this branch's head 61a8e5f is an ancestor of main). `verify_claim` is wired in `tool-dispatch.ts`, classified `conclusion` in `tool-contract.ts`, and gated to the opt-in `verify` preset; its full test suite (claim-verification, tool-contract, adversarial) passes on main. GitHub couldn't auto-mark this MERGED because its base was the #165 branch rather than `main`. Closing as included.

PR #172, opened 2026-06-20
feat(analyzer): link HTML to its external JS/CSS as dependency edges

laurentftech:
## Problem

An `.html` file had **no edges to the assets it loads**. `<script src="app.js">` and `<link rel="stylesheet" href="style.css">` were unparsed, so:

- `get_file_dependencies("index.html")` showed nothing — every page looked isolated.
- `analyze_impact("style.css")` / `analyze_impact("app.js")` never listed the HTML pages consuming them — the blast radius of editing a shared stylesheet/script was invisible.
- The dependency graph treated each HTML page as a leaf, distorting file-level topology.

Distinct from #171 (inline `<script>` → call-graph *function nodes*); this adds *file→file dependency edges*.

## Approach — reuse the existing edge machinery

`DependencyGraphBuilder.buildEdges` already creates a file→file edge for every `FileAnalysis.imports` entry that `resolveImport` resolves to a graph node, and `.css`/`.js` files are already nodes. The only gap: HTML produced no `imports`.

- **`parseHtmlAssetImports(content)`** — emits `ImportInfo[]` for `<script src=…>` and `<link rel="stylesheet" href=…>`.
- **`normalizeAssetHref`** — drops external URLs (`http(s)://`, `//`), `data:`/`mailto:`/`tel:`/`javascript:`, `#` anchors, root-absolute `/…` (web-root, out of scope); strips `?`/`#`; prefixes bare `app.js` → `./app.js` so it resolves against the document directory.
- **Dispatch** — `getFileType` gains an `.html`/`.htm` arm; `parseFile` routes to the new parser. `buildEdges` + `resolveImport` then create the edges unchanged (the exact-path resolver candidate handles `.css`/`.js`).
- **`assetKind?: 'script' | 'stylesheet'`** — additive field on `DependencyEdge`, carried from the import, so `get_file_dependencies` can label "loads asset" vs "imports symbols".

## Safety / robustness (review-hardened)

- **Traversal-safe.** A `../../../x` href can only ever edge to a file already in the repo map — the `fileSet.has(resolvedPath)` gate blocks any out-of-repo target. Root-absolute `/…` is excluded before resolution.
- **Comments ignored.** HTML comments are blanked (same-length whitespace, line numbers preserved) so commented-out `<script>`/`<link>` tags don't produce phantom edges.
- **`rel` matching** is left-anchored (no `data-rel` false positive) and order-independent (`rel="preload stylesheet"` is detected).
- **No ReDoS** — bounded `[^>]` scans, no nested quantifiers.

## Out of scope (documented)

- Root-absolute / web-root href resolution (needs a configured web root).
- ``, `<a href>`, `<source>`/`srcset`, CSS `@import`, framework template asset refs.
- A `>` inside an attribute value before `src`/`href` (rare) truncates the tag match — known limitation.

## Tests

`html-asset-imports.test.ts` (9): script src + stylesheet link, external/data/anchor/root-absolute excluded, `rel=preload/icon/manifest` excluded, bare-href normalization, query/fragment strip, quote style + attribute order, line numbers, **commented-out tags ignored**, **reversed multi-token `rel`**, **`data-rel` decoy**, **multiline tag**.

`html-asset-edges.integration.test.ts` (2): real `RepositoryMapper` + `DependencyGraphBuilder` → `index.html → app.js` + `→ style.css` edges with `assetKind` and raised stylesheet in-degree; CDN-only page → no edges.

Full suite green except one pre-existing `memory-invariant` timeout (unrelated — fails identically on the clean base).

A `code-reviewer` pass returned APPROVE; its MEDIUM findings (commented-out tags, `data-rel`) and the reversed-`rel` LOW are fixed here with regression tests.

### DIFF
--- a/openspec/changes/add-html-asset-dependency-edges/proposal.md
+++ b/openspec/changes/add-html-asset-dependency-edges/proposal.md
@@ -1,0 +1,77 @@
+# HTML asset-dependency edges: link `.html` to its external JS/CSS
+
+> Status: DRAFT (2026-06-20). Decision `b555b680` recorded before any code.
+> Distinct from the inline-script call graph (`5b38bad2`): that adds *function nodes* for inline
+> `<script>` JS; this adds *file→file dependency edges* from an HTML page to the external assets it
+> references. Distinct again from the literal-text line index (`fd256fde`).
+
+## Why
+
+An `.html` file has **no edges to the assets it loads**. `<script src="app.js">` and
+`<link rel="stylesheet" href="style.css">` are unparsed, so:
+
+- `get_file_dependencies("index.html")` shows nothing — the page looks isolated.
+- `analyze_impact("style.css")` / `analyze_impact("app.js")` never lists the HTML pages that consume
+  them, so the blast radius of editing a shared stylesheet or script is invisible.
+- The dependency graph treats every HTML page as a leaf with no out-edges, distorting the file-level
+  topology for any project that wires its front-end through HTML.
+
+The machinery to fix this already exists. `DependencyGraphBuilder.buildEdges` creates a file→file edge
+for every entry in a file's `FileAnalysis.imports` that `resolveImport` can resolve to a file already in
+the graph (`dependency-graph.ts:331`). `.css` and `.js` files are already walked and present as nodes.
+The only missing piece is that HTML files produce **no** `imports`.
+
+## What changes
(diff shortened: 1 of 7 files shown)


clay-good:
Thank you @laurentftech  ❤️ 🚀

PR #173, opened 2026-06-20
feat(watcher): full watch-mode reconciliation — dependency graph, HTML, create/delete

laurentftech:
Makes `mcp --watch` reconcile **modify, create, and delete across every lane** — closing the watch-mode staleness gaps. Commits are stacked smallest-first.

## 1. Live dependency-graph edges in watch mode
The watcher kept signatures, the call-graph EdgeStore, the text-line index, and the vector index live — but **never touched `dependency-graph.json`** (what `get_file_dependencies` reads). So **any** import edit went stale until a full `analyze`. Cheap fix: extract `computeFileImportEdges` from `buildEdges` (shared, no drift) and patch the changed file's edges + degrees in `updateDependencyGraph` (atomic tmp+rename). HTTP/call-synthesized edges preserved; global metrics (pageRank/clusters) deferred to full analyze.

## 2. HTML live in watch mode
The watcher ignored HTML (`unknown` language, not in `SOURCE_EXTENSIONS`), so the three HTML features (text-line, inline-`<script>` call-graph, `<script src>`/`<link>` asset edges) only refreshed on full analyze. Admit `.html`/`.htm`; **`buildGraphSubset` blanks HTML** (offset-preserving) so an edit refreshes inline-script nodes instead of the swap wiping them. Text-lines + asset edges then refresh for free.

## 3. Create & delete reconciliation
The watcher only handled `'change'`. Now:
- **`'add'`** → the change pipeline; `updateDependencyGraph` adds a node (+ outgoing edges) for a not-yet-known file.
- **`'unlink'`** → `handleDeletions` removes the file from **every** lane: call graph (edges both directions + nodes + cfg + classes), signatures, text-line index, vector index, dependency graph (node + touching edges, degrees recomputed, atomic write).
- Coalescing: deletions share the change debounce via `pendingDeletions`; a flush does deletions first then changes; a re-create supersedes a pending delete and vice-versa (and `stop()` drains both on shutdown).

## Scope (by design, documented)
Global dep-graph metrics, new-file *incoming* edges, `.vue`/`.svelte`, and `.css` → full analyze. Watch-ignored paths (`*.test.ts`, build dirs) never fire events.

## Tests
- `mcp-watcher.integration.test.ts`: import re-point, call-edge preservation, new-file node add, deletion node/edge removal, **supersession** (delete→recreate ⇒ present, recreate→delete ⇒ removed), HTML `<script src>` re-point.
- `mcp-watcher-html.test.ts`: `buildGraphSubset` HTML blanking (node-wipe regression guard).
- `dependency-graph-watch.e2e.integration.test.ts`: full **analyze → watch → get_file_dependencies** loop for modify, create, delete, and HTML.
- Existing dependency-graph + import-parser + watcher suites pass unchanged. Full suite otherwise green; only failures are the pre-existing `memory-invariant` property tests (60s timeouts, load-sensitive, fail identically on the clean base).

Three `code-reviewer` passes: dep-graph (fixed HIGH torn-read → tmp+rename), HTML (APPROVE), create/delete (fixed MEDIUM shutdown-drains-deletions; the test-file-leak finding was moot since test files are watch-ignored).

### DIFF
--- a/openspec/changes/add-watch-incremental-dependency-graph/proposal.md
+++ b/openspec/changes/add-watch-incremental-dependency-graph/proposal.md
@@ -1,0 +1,69 @@
+# Live dependency-graph edges in watch mode
+
+> Status: DRAFT (2026-06-20).
+> Closes a systemic watch-mode staleness: the call graph (function→function) is kept live by the
+> watcher, but the **dependency graph** (file→file imports) was never updated incrementally — it froze
+> until a full `analyze`. This affects every language's import edits, not just HTML assets.
+
+## Why
+
+`get_file_dependencies` reads the static `dependency-graph.json` artifact (`graph.ts`). The MCP watcher
+incrementally maintains signatures, the call-graph EdgeStore, the text-line index, and the vector index
+— but **never touches `dependency-graph.json`**. So in `mcp --watch`, the moment an import is added,
+removed, or re-pointed, the file→file dependency view goes stale until the next full `analyze`:
+
+- `get_file_dependencies("x")` returns the *old* imports.
+- file-level in/out-degree, and anything reading the dep-graph artifact, drift.
+
+(The `serve` daemon hides this by scheduling a debounced **full** re-analyze via `onBatchFlushed`, but
+that is O(repo); plain `mcp --watch` has no such fallback.)
+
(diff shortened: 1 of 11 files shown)


clay-good:
Thank you @laurentftech ❤️

PR #264, opened 2026-07-21
fix: normalize Windows backslash paths before splitting on '/'

zoroneo:
## Problem

On Windows, file paths use backslash (`\`) as the separator, but several path-matching functions split on `/` without normalizing first. This causes all path-based domain inference to produce a single segment (e.g. `"src\auth\service.ts"` stays as one token), so no domain matches and every file returns `'misc'`.

This makes `openlore verify` fail with `"No suitable verification candidates found"` on Windows because:

1. `inferDomain()` in the verification engine splits on `/` -> gets one segment -> returns `'misc'` -> all files filtered out
2. `inferDomainFromPath()` in the spec mapper has the same issue
3. `inferDomains()` in the repository mapper can't cluster files by domain on Windows

## Fix

Add `.replace(/\\/g, '/')` before every `.split('/')` that operates on file paths from the filesystem (which use OS-native separators):

| File | Method | Change |
|------|--------|--------|
| `src/core/verifier/verification-engine.ts` | `inferDomain()` | `filePath.split('/')` -> `filePath.replace(/\\/g, '/').split('/')` |
| `src/core/verifier/verification-engine.ts` | `normalizeImport()` | Same pattern (handles C# import paths on Windows) |
| `src/core/drift/spec-mapper.ts` | `inferDomainFromPath()` | Same pattern |
| `src/core/analyzer/repository-mapper.ts` | `inferDomains()` | Same pattern |

Other files in the codebase (`github-actions.ts`, `docker.ts`, `classify-yaml.ts`, `refresh-stories.ts`) already use this pattern (`path.replace(/\\/g, '/').split('/')`). This fix extends the same pattern to the remaining locations.

## Testing

- `npx vitest run src/core/verifier/` -- **41/41 passed**
- `npx vitest run src/core/drift/spec-mapper.test.ts` -- 89/90 passed (1 pre-existing failure on Windows: `relative()` returns backslashes in `adrPath` test assertion)
- `npm run build` -- compiles cleanly (TS5101 deprecation is pre-existing)

## Verification on a real Windows repo

After this fix, `openlore verify` on a 3,300+ file C# WinForms project correctly:
- Loads 1448 file-to-domain mappings
- Selects 30 verification candidates across 14 domains
- Runs LLM-based verification (28/30 files verified)

### DIFF
--- a/src/core/analyzer/repository-mapper.ts
+++ b/src/core/analyzer/repository-mapper.ts
@@ -476,7 +476,7 @@ function inferDomains(files: ScoredFile[]): Record<string, ScoredFile[]> {
     // (src/main/java/com/example/inventory/Foo.java) at the business package
     // ("inventory") instead of collapsing every source file into the org root
     // ("com"/"springframework"). See issue #138.
-    const dirParts = file.path.split('/').slice(0, -1);
+    const dirParts = file.path.replace(/\\/g, '/').split('/').slice(0, -1);
     const domain = deriveDomainFromPath(dirParts);
     if (domain) {
       if (!domains[domain]) {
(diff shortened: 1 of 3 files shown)


clay-good:
Thank you @zoroneo ❤️

PR #268, opened 2026-07-25
perf(analyzer): run Pass-1 extraction on a worker pool, byte-identically

clay-good:
**Status:** Ready for review. Full suite green (321 files / 6146 tests), lint + typecheck clean, byte-equality oracle passes end to end, and two rounds of adversarial review are folded in.

Implements change `optimize-parallel-extraction-pool` (proposal 1 of the competitive-substrate sweep).

## What was missing

`openlore analyze`'s dominant cost is Pass 1 — per-file tree-sitter parsing and fact extraction — and it ran as a strictly serial `for…of` on the main thread. On this repo that is **14.3s of a 20.2s call-graph build (71%)**, with every other core idle. Parsing is embarrassingly parallel per file (no file's extraction reads another file's state), and there was no worker anywhere in the pipeline.

Parse throughput is the substrate's admission price: "re-analyze after significant refactors", zero-interaction onboarding, and the self-healing rebuild are all bounded below by how fast analyze runs.

## What it does

A fixed-size `worker_threads` lane for Pass 1, sized `min(availableParallelism() - 1, remaining process budget, fileCount)` and started only above `EXTRACTION_POOL_MIN_FILES`.

The design is built around one property — **the pool must make extraction faster without making it different**:

- **No extraction logic in the pool.** Workers call the same `dispatchFileExtract` the serial lane calls. The pool module is scheduling and failure handling only, so the two lanes cannot drift.
- **Index-keyed slots, never completion order.** Pass 1 is now an EXTRACT step (pooled or serial) followed by a MERGE step that walks `files` by index. Every determinism property downstream — which node wins an id collision, the raw-edge sequence — is a property of the merge loop, not of the lane.
- **The serial lane stays the reference implementation** and is also the fallback executor.

**Deviation from the proposal, deliberate:** workers receive the file's *content*, not just its path. `build`'s input content is authoritative and is not always what is on disk — HTML pages arrive inline-script-blanked, the incremental path passes in-memory content, and test builds pass synthetic files with no on-disk twin. A worker-side re-read would change facts, not just I/O. The proposal and spec delta carry that rationale.

## The interesting failure mode: a worker that goes blind

The extractors report an unavailable grammar by returning an **empty result**, not by throwing. A worker whose grammar for some language failed to load *in its thread* would therefore report every file of that language as containing nothing, and the merge would record that as "no symbols here" — precisely the silent loss the parse-health work exists to prevent. A startup probe cannot close this: it proves one grammar, and the other ~20 load lazily and independently per thread.

So **a worker's silence is not trusted until that worker has demonstrably extracted that language**. An empty result for an unproven language is re-checked on the main thread; if the main thread then finds facts, that is disclosed as a lane defect rather than recorded as an empty file. Measured cost on a healthy run of this repo: **zero re-checks** — a real parse of a function-less file still yields style counters, so ordinary empty files are never mistaken for silence. The bounded cost lands only on languages with no style tally, and only until that worker's first real result.

The startup probe still runs as a fast-fail, now in a language the build actually contains (every grammar is an optional dependency, so probing TypeScript in a Python repo could disable the pool over a grammar that repo never needed).

## Fail-soft ladder

Every rung is disclosed on `CallGraphResult.extractionLane` and surfaced by the CLI; none of it changes a fact.

| Failure | Response |
|---|---|
| A worker dies mid-file | that file is re-extracted on the main thread |
| An extractor *throws* inside a worker | recorded as that file's parse failure — identical to serial, no retry |
| A worker returns nothing for an unproven language | re-checked on the main thread; a disagreement is disclosed as a lane defect |
| A worker replies off-protocol, or stops answering | retired on a deadline; its file falls to the main thread |
| No worker comes up healthy / no worker entry / `OPENLORE_NO_WORKERS=1` | wholesale serial lane (and the process remembers, so later builds don't re-pay the stall) |

**Nothing on this path writes to stdout.** `openlore mcp` speaks JSON-RPC over stdout and runs builds in-process, and a worker thread inherits none of the parent's console patching. Workers keep their stdout off the parent's (drained; stderr stays inherited, since it is not a protocol channel and losing a crash trace is worse), and the lane note is returned on the build result for the CLI to render instead of being logged from the builder.

**Worker count is bounded per process, not per build** — the daemon runs a background self-heal rebuild alongside tool-call builds, and each would otherwise plan a full pool and double the resident isolate count.

## Proof it works

**Measured** (this repo, macOS, 10 cores, pool size 8):

| | serial | pooled |
|---|---|---|
| call-graph build | 18.9s | **7.2s** (2.6×) |
| `openlore analyze --no-embed` | 24.7s | **12.6s** (2.0×) |

**Byte-equality oracle** — `openlore analyze` on a clean checkout, pooled and with `OPENLORE_NO_WORKERS=1`. Every content artifact is byte-identical: `llm-context.json`, `repo-structure.json`, `dependency-graph.json`, `style-fingerprint.json`, `parse-health.json`, every inventory, `CODEBASE.md`, `ARCHITECTURE.md`. Four artifacts differ — `SUMMARY.md`, `fingerprint.json`, `refactor-priorities.json`, `vector-index-meta.json` — in a wall-clock timestamp only, and they differ identically between two *same-lane* runs, so that is pre-existing and not lane-dependent.

**MCP stdio purity** — drove `openlore mcp` over stdio through a cold-start index build (which runs the pool in-process) plus an `orient` tool call: 3 stdout lines, all valid JSON-RPC frames, all requests answered, build noise on stderr.

**Tests** (26 across two files):
- *Order independence*: a stub pool that answers later files **first** — asserting the completion order really was inverted — still yields a byte-identical serialized graph, including a colliding-symbol corpus where merge order alone decides which node survives.
- *Blind worker*: a worker that reports nothing for a language still yields a graph byte-identical to serial, and its defect is disclosed. A function-less file costs zero re-checks; a language with no style counters is re-checked only until the worker proves it; a language with no extractor is a deterministic answer, not a silence.
- *Worker crash*: one worker dies mid-file, and separately **all** workers die. Every file accounted for, facts identical to serial, fallback disclosed.
- *Protocol and liveness*: an off-protocol reply retires the worker instead of hanging the pass; a worker that never reports ready is dropped on its deadline (driven with fake timers, so it cannot quietly become a slow test).
- *Process budget*: concurrent builds share one worker budget, and it is returned when they finish.
- *Real threads*: actual `worker_threads` spawn, probe, extract, and match serial byte for byte — including the Lua and Dart **WASM** grammars, whose emscripten heaps must be isolated per thread, and a per-language sweep over Python/Go/Rust/Ruby/Java/C++/Swift/JavaScript (a grammar that loads on the main thread but not in a worker returns *empty* rather than throwing, so one language cannot speak for the rest).
- *Shipped entry*: `dist/core/analyzer/extraction-worker.js` — the file the npm package ships, which no other test loads because vitest always resolves the `.ts` entry — starts and reports ready, and says so loudly if the checkout is unbuilt rather than passing quietly.

## Notes / nits

- **CI's Lint & Type Check job fails on `npm audit --audit-level=high`, which runs before lint and is red on `main` itself** (the `sharp`/libvips advisory chain, handled separately on `fix/security-dependency-alerts`). Nothing here touches dependencies; `npm run lint` and `npm run typecheck` are clean.
- The suite runs with `OPENLORE_NO_WORKERS=1` (both vitest configs) so the ~320 other test files never pay for thread spawn. The pool is covered deliberately instead: a stub lane for ordering/failure semantics, and one file that clears the flag to exercise real threads.
- Scope is Pass 1 only. Global passes (resolution, synthesis, communities) stay single-threaded — they are cross-file and not the dominant cost; their pass-count reduction is the sibling `optimize-analyze-pipeline-passes`.
- No tool, no output shape, and no conclusion changed — so there is no MCP↔Pi parity work here.
- `extraction-worker.ts` is reached only through a runtime path resolution, so OpenLore's own `find_dead_code` / `report_coverage_gaps` will report its functions as unreachable. That is a true statement about the static graph, not a bug in either.
- `OPENLORE_NO_WORKERS=1` is the documented escape hatch (now in `docs/configuration.md`), and the serial lane is exercised on every run of the suite, so the cold path cannot rot.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,6 +106,14 @@ jobs:
         # diff, and this checks the real `npm pack` manifest rather than the intent.
         run: npm run audit:packlist
 
+      # The Pass-1 extraction worker is loaded by runtime path resolution, and under vitest
+      # the .ts entry is always the one resolved — so the COMPILED entry that the npm package
+      # actually ships is exercised by no other job. A compile or packaging regression there
+      # degrades every installed user to the serial lane silently, so it is checked here,
+      # where dist/ exists (change: optimize-parallel-extraction-pool).
+      - name: Verify the compiled extraction worker
+        run: npx vitest run src/core/analyzer/extraction-pool-threads.test.ts
+
       - name: Upload build artifacts
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -44,6 +44,7 @@
 | `OPENLORE_NO_AUTO_ANALYZE` | -- | Disable the MCP server's cold-start self-bootstrap (no background index build on first run) |
 | `OPENLORE_NO_UPDATE_NOTIFIER` | -- | Silence the passive "update available" banner (`NO_UPDATE_NOTIFIER` is also honored) |
 | `OPENLORE_SKIP_POSTINSTALL` | -- | Suppress the post-install next-step hint |
+| `OPENLORE_NO_WORKERS` | `analyze` | Run per-file extraction on a single thread instead of the worker pool. Both lanes produce byte-identical analysis output, so this only costs wall-clock — set it to isolate a worker-related problem, or in an environment where extra threads are unwelcome |
 
 > The `EMBED_*` variables configure the **remote** embedding provider only. For on-device embeddings with no endpoint or key, run `openlore embed --local` (or set `embedding.provider: "local"` in `.openlore/config.json`). Keyword (BM25) search is the first-class default and needs none of these. See [docs/semantic-search.md](semantic-search.md#retrieval-modes) for the full embedding/retrieval-mode reference.
 
--- a/openspec/changes/COMPETITIVE-SUBSTRATE-2026-07.md
+++ b/openspec/changes/COMPETITIVE-SUBSTRATE-2026-07.md
@@ -1,0 +1,68 @@
+# Change set: competitive substrate sweep 2026-07 — faster than the field, more useful than the graph
+
+This directory gained 6 change proposals on 2026-07-23, produced by a three-track research sweep
+of the code-intelligence-for-agents field (the current graph-index entrants, the fast-indexing
+engineering literature, and the published evidence on what actually improves agent success),
+diffed against the existing 141-proposal backlog so every item here is net-new whitespace. Per
+project policy, no competing product is named in any proposal; concepts are credited to the
+field or to academic sources.
+
+## What the research established
+
+**Positioning is validated.** Every breakout tool of 2025–26 in this category is local-first
+with embedded storage over MCP; the server-dependent graph-DB tools stalled on ops friction, and
+one team's own tracker records regretting a move OFF an embedded SQLite store. Conclusion-shaped
+one-call answers are the convergent adoption winner (published measurement: agents underuse
+graph tools that demand query formulation and fall back to grep — https://arxiv.org/pdf/2602.20048),
+which is OpenLore's shipped doctrine. Nobody else has symbol-anchored self-invalidating memory,
+spec drift, governance findings, or honest-boundary disclosure — the governance face is
+uncontested.
+
+**The speed bar is set by three mechanisms** the field converged on, all absent here:
+content-hash-keyed per-file memoization (re-index cost proportional to the diff), parallel
+per-file parsing, and resolve-at-index-time serving (queries as lookups, not per-call
+traversals). The evidence-backed usefulness gaps are span-level localization (agents hit the
+right file 60–70% of the time but the right lines only 14–19%; evidence-per-token correlates
+r≈0.95 with repair success — https://arxiv.org/html/2606.07297v1) and deterministic per-edit
+verification loops (+21–32 pt correctness in feedback-loop studies —
+https://arxiv.org/html/2504.06939v2).
+
+## The six proposals
+
+| # | Change | Track | One line |
+|---|--------|-------|----------|
+| 1 | `optimize-parallel-extraction-pool` | speed | Pass 1 parses serially on one core (`call-graph.ts:3979-4034`, zero workers repo-wide); a worker pool with input-order merge ≈ core-count× on the dominant phase, byte-identical output |
+| 2 | `optimize-hash-keyed-analyze` | speed | Batch analyze is skip-all or re-parse-all (`analyze.ts:411-412`); `file_hashes` exists but only the watcher reads it — per-file fact memoization makes analyze O(diff) |
+| 3 | `optimize-reachability-precompute` | speed | Every flagship conclusion re-runs BFS over per-call-built adjacency (no condensation/topo/closure exists); precompute at analyze, serve as lookups |
+| 4 | `add-span-precise-conclusions` | usefulness | The line-precise substrate (call-site lines, def-use lines) is persisted and then dropped at the conclusion boundary; surface it + a `focus` slice on `get_function_body` |
+| 5 | `add-edit-loop-breakage-verdict` | usefulness | The watcher knows about breakage at patch time and says nothing; derive a provable-only per-edit verdict (broken refs, arity, imports, reaching tests), hook-deliverable, advisory |
+| 6 | `refine-first-run-partial-serving` | onboarding | First build serves "no index found" for minutes despite significance-ordered (hubs-first) processing; flush partials + completeness receipt, extend the stale-serving contract to absent |
+
+## How they compose
+
+1+2 multiply (parallelize only the diff); 3 is the serving-side floor that keeps 5 sub-second at
+scale; 4 and 5 are the two evidence-ranked usefulness levers (localization depth, verification
+loop); 6 converts 1+2's speed into the first-contact experience. Cross-references to the
+existing backlog are named inside each proposal (`optimize-analyze-pipeline-passes`,
+`optimize-serving-hot-path-caches`, `optimize-incremental-and-coldstart-scale`,
+`add-incremental-early-cutoff`, `add-symbol-content-hashes`, `add-agent-loop-enforcement-hook`,
+`refine-orient-context-budgeting`) — none is re-proposed.
+
+## What was considered and deliberately not proposed
+
+- **Trigram/sparse-ngram literal index** — the literal-text line index shipped
+  (`text-line-index.ts`) covers the need at current scale; a posting-list sidecar is the
+  monorepo-era follow-up, not now.
+- **Reachability interval labeling (GRAIL/FERRARI class)** — condensation walks suffice below
+  hundreds of thousands of nodes; labeling adds an invalidation liability (noted as an explicit
+  non-goal in proposal 3).
+- **LLM-summarized "purpose" embeddings** — the one LLM-at-index-time idea with field evidence,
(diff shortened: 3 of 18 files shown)


clay-good:
CI note: the **Lint & Type Check** job fails on `npm audit --audit-level=high` — that step runs before lint and is failing on `main` itself (runs for `9020cff`, `e47ad58`, `6572a75` are all red for the same reason; it's the `sharp`/libvips advisory chain being handled on `fix/security-dependency-alerts`). Nothing in this PR touches dependencies. `npm run lint` and `npm run typecheck` are clean locally.

The **Unit Tests** failure on the first push was mine and is fixed in `74e1934`: the dead-worker test ran three 40-file builds of real tree-sitter extraction and blew vitest's 5s default on the slower runner. Corpora are now the smallest that still exercise each hazard, every pool test has an explicit 30s budget, and the thread suites now assert the pooled lane really was pooled — `OPENLORE_NO_WORKERS` silently forces serial, which would have made those byte comparisons vacuous serial-vs-serial passes.

clay-good:
Hardening round is in (`9e62a64`, `210aa44`). Two independent adversarial reviewers found three real problems, none of them in the ordering logic:

1. **A worker could go blind and be believed.** The startup probe covers one grammar; the other ~20 load lazily and independently per thread, and an unavailable grammar returns an *empty* result rather than throwing — so a worker whose grammar for some language failed to load would report every file of that language as containing nothing, and the merge would record that as "no symbols here". A worker's silence is now untrusted until that worker has demonstrably extracted that language. Measured cost on a healthy run of this repo: **zero re-checks**, because a real parse of a function-less file still yields style counters.
2. **Two paths could hang analyze forever** — an off-protocol reply left its request armed permanently, and a worker wedged in native parse code emits neither reply nor exit. Both now have deadlines that retire the worker and hand its file to the main thread.
3. **stdout is a protocol channel.** `openlore mcp` speaks JSON-RPC over stdout and runs builds in-process; a worker thread inherits none of the parent's console patching. Workers keep their stdout off the parent's, and the lane note is returned on the build result for the CLI rather than logged from the builder. Verified end to end: drove the stdio server through a cold-start pooled build plus an `orient` call — 3 stdout lines, all valid JSON-RPC, build noise on stderr.

Also added a process-wide worker cap (the daemon runs a self-heal rebuild alongside tool-call builds; each would otherwise plan a full pool) and a test for the *compiled* worker entry, which no other test loads because vitest always resolves the `.ts` one.

Numbers after the fixes: call-graph build 18.9s → 7.2s, full `analyze --no-embed` 24.7s → 12.6s, artifacts still byte-identical between lanes.

clay-good:
CI accounting, to be precise about what is and isn't green here:

| Job | Status |
|---|---|
| Build | ✅ pass |
| Unit Tests | see below |
| Lint & Type Check | ❌ — fails on `npm audit --audit-level=high`, which runs *before* lint in that job |

The audit failure is not from this branch. It is the same failure `main` has had since `9020cff` (the `sharp`/libvips advisory chain), and **#267 fixes it** — that PR is fully green including Lint & Type Check. Once #267 lands, a re-run here goes green with no change to this branch. `npm run lint` and `npm run typecheck` both pass locally on this branch.

On unit tests: two earlier runs surfaced failures in `mcp-watcher-incremental` (`G2/G5`, `G3`) and `decisions` (`C9`). Neither is this change — the watcher tests assert against a fixed `setTimeout(200)` with a 30ms debounce, and C9 races two concurrent CAS writers — but the pool suite does real tree-sitter extraction and was adding enough CPU pressure on a 4-core runner to tip them. `30a4bdb` cuts that suite from ~9.4s to ~4s (the scheduling-independence test dominated; it asserts the same two things at 8 builds over 12 files instead of 12 over 24). Flagging those two tests as genuinely load-sensitive rather than touching them here.

clay-good:
✅ **Unit Tests pass** on `30a4bdb` — the CPU trim resolved the collateral flakiness in the watcher and decisions tests. Build passes too.

Final CI state on this branch: Build ✅ · Unit Tests ✅ · Lint & Type Check ❌ (`npm audit`, red on `main` itself, fixed by #267 — that PR is fully green including this job). Nothing further to do here until #267 lands.

clay-good:
Converged. An independent third review pass verdict: **CONVERGED — ship it**, after two blockers it found were fixed. Both were tests that did not test what their names claimed:

1. **The shipped-entry guard never ran.** It resolved `join(__dirname, 'extraction-worker.js')` — a sibling `.js` that tsc never emits (`src/` → `dist/`). In a fully built checkout it printed "dist entry absent", passed in 1ms, and guarded nothing, while tasks.md claimed it proved the compiled worker starts. Repointed at repo-root `dist/`, wired into the CI **Build** job (the only job where `dist/` exists), and verified it *fails* against a deliberately corrupted compiled entry.
2. **The "extractor throws inside a worker" test asserted nothing about that.** It injected the throwing extractor as `serialExtract`, which a healthy stub worker never calls — so no `failed` reply was ever produced and it would have passed with the worker error path deleted outright. Deleted; the behavior is covered by the proven-worker error test.

Plus, from the same pass: a table-driven guard over the worker's `PROBES` snippets (one that quietly stopped yielding a node+edge would disable the pool process-wide with no failing test), removal of a write-only field and the generic it forced, and honesty fixes to the disclosure text and the "byte-identical" comments (a worker thread's larger default stack means a pathological deep-recursion parse can fail on one lane and not the other — a different *file* failing to parse, disclosed via parse health either way).

One hazard I closed on my own after that review: importing the worker entry for its `PROBES` table is only safe because vitest 4 defaults to the `forks` pool. Under a thread pool the import would patch the shared logger and post a stray frame on the runner's RPC channel. The pool now sends a sentinel in `workerData` and the entry stays inert without it — verified both directions against the compiled entry.

**On the intermittent `serve.test.ts` failure** (`ENOTEMPTY … rmdir openlore-serve-reset-*`): I reproduced it on `origin/main` in a git worktree, 1 run in 6, in a file this change does not touch. Pre-existing temp-dir cleanup race, not this PR.

clay-good:
CI's Unit Tests job failed four tests in `mcp-watcher-incremental.test.ts` — and they are exactly the four that slept a fixed 150–200ms after enqueuing, then asserted the debounced flush had already happened. Fixed in `c332337` by polling for the outcome each test already asserts (the summary line, the prime call, the persisted file) with a 5s ceiling. Assertions unchanged; only the guess about machine speed is gone.

This is a file unrelated to the pool, and I'd normally leave it alone — but the load is mine to account for: `main`'s CI passes this file, and the pool suite added here does real tree-sitter extraction on a 4-core runner.

Honest caveat: I could **not** reproduce the failure locally, even with every core saturated — this box has 10 cores, CI has 4 running ~320 test files in parallel. The diagnosis rests on the exact correspondence between the four sleeps and the four failures, and on the replacement being strictly weaker in its assumptions rather than differently timed. If CI still fails after this, that correspondence is wrong and I'll dig further rather than keep tuning timings.

clay-good:
Correction to my last comment: I attributed the watcher-test failures to load from the pool suite. That was wrong, or at least incomplete.

Working on an unrelated branch (#269 — markdown changes only, no pool code at all), `G2/G5` failed **2 of 3** local runs. So those tests are fragile on their own; the pool suite raises the odds but is not the cause.

The fix in `c332337` stands — polling for the outcome instead of sleeping a fixed 200ms is correct either way, and CI's Unit Tests went green with it. But the commit message says the pool suite "is what tipped them over", and that overstates what I actually knew. Worth correcting since the whole point of that commit was to be honest about a diagnosis I could not reproduce.

PR #284, opened 2026-07-25
fix(security): fix CodeQL findings, scope TLS per-request, and tune the analysis scope

clay-good:
## Status

LGTM. All 7 checks green. 324 files / 6,204 tests.

Two parts: the CodeQL triage, and a follow-up adversarial pass that found two defects **in this PR's own first commit** plus the TLS scoping.

---

# Part 1 — CodeQL triage

The first baseline was **458 alerts**. All triaged.

## Why 80% was noise

**368 of 458 were in test files**, and one rule — `js/insecure-temporary-file` (379) — accounts for most: fixtures legitimately write to `tmpdir()`.

That taint then flowed **into production sinks**: 30 alerts landed on production `writeFile` calls whose path came from a *test caller*. `gitignore-manager.ts` was reported despite containing **no reference to `tmpdir()` at all**. So the noise was misattributing findings to files that don't have the problem.

`paths-ignore` now excludes `*.test.ts` / `*.spec.ts`, justified on **scope**: tests aren't shipped (the `files` allowlist omits them, `audit:packlist` enforces it) and aren't reachable outside the runner. The config records the policy: **exclude non-shipped code, dismiss individual accepted findings with a reason, never silence a rule.**

## Four real fixes

| Rule | The actual failure |
|---|---|
| `js/regex-injection` | `functionName` is a caller-supplied MCP argument spliced into `\b<name>\s*[(<]`. Unescaped, `foo(.*)\|bar` matches `bar(` — `get_function_body` returns **a different function than asked for**. |
| `js/incomplete-sanitization` | DOT output escaped `"` but not `\`, so a trailing backslash escaped its own closing quote. |
| `js/prototype-pollution-utility` | `setPath` walks a dotted path into a plain object; a `__proto__` segment assigns onto `Object.prototype`. |
| `js/remote-property-injection` | `directoryCounts` is keyed by repository directory names. |

Two went past what CodeQL flagged: the DOT fix also covers node/edge ids (interpolated with *no* escaping — same hole, unflagged), and the pollution guard covers `deletePath` too.

---

# Part 2 — Adversarial pass

## Two defects in this PR's own first commit

**1. The pollution guard turned a vulnerability into a crash.** `setPath`/`deletePath` throwing is right for the entries *we* pass (internal constants). But `isHandEdited` and `removeManaged` feed them `_openlore.paths` — read back off the **user's config file**. A hand-edited file made `openlore install` and `--uninstall` throw instead of proceeding. Reproduced, then fixed: file-derived paths are filtered and ignored (OpenLore never writes such a path, so it can't describe anything we manage); our own paths still throw.

**2. `escapeDotString` was still incomplete.** It escaped `\` and `"` but not control characters — and the emitter joins statements with newlines, so a file name containing a newline (legal on Linux/macOS) split the statement and the tail parsed as new DOT. Demonstrated, then closed.

## One more instance of the same class

`repository-mapper`'s `clusterFiles` and `inferDomains` build the same repo-keyed lookup tables, and **CodeQL did not flag them**. For a directory named `__proto__`, `!map[key]` reads `Object.prototype` (truthy), skips the array init, then throws on `.push`. Both are now prototype-less — verified against a real repository containing a `__proto__` directory.

---

# Part 3 — TLS is now scoped per request

`NODE_TLS_REJECT_UNAUTHORIZED = '0'` was set at start-up and **left set**. From the first LLM call onward, every https connection in the process was unverified — not just the endpoint the user opted out for. In `openlore mcp`, that meant the rest of the session.

New `tls-scope.ts` centralises it. `allowInsecureTls()` records the opt-in without disabling anything; `withRelaxedTls()` sets the variable immediately before a request and restores it immediately after, refcounted so a concurrent request can't restore it early. All 8 fetch sites are wrapped, and **no module writes the variable directly** any more.

### Verified against a real handshake, not assumed

The design rests on two Node behaviours, both asserted in tests using a self-signed cert **generated at run time** — not committed, since a checked-in key is exactly what push protection exists to reject:

```
self-signed rejected outside the scope   ✓
accepted inside the scope                ✓
verification restored while body streams ✓  (handshake already done)
new request rejected after               ✓  (relaxation did not persist)
```

Plus the state machine: no-op until opt-in, restores on throw, restores a pre-existing value verbatim, and keeps the scope open until the **last** concurrent request finishes.

### The limit, stated plainly

The variable is still process-global while a scope is open, so an unrelated request in flight during that window is also unverified. **The window is one request instead of the process lifetime** — a large reduction, not elimination.

Removing it entirely needs per-request TLS options, which Node's built-in `fetch` does not expose. I tested the obvious alternative: an npm `undici` `Agent` passed as `dispatcher` is **rejected with `UND_ERR_INVALID_ARG`**, because the built-in fetch is a separate undici instance. Undici's *own* `fetch` does honour it (also verified), so that route means adopting `undici` as a dependency and switching these call sites to its fetch — a behaviour-affecting change that belongs in its own PR.

`--quiet` stays honest: the helper takes `announce: false` so the CLI's colorized, quiet-aware notice isn't duplicated and nothing is written in quiet mode. Verified end to end — `--insecure doctor` prints exactly one notice, `--insecure --quiet doctor` prints none.

---

## Remaining production alerts — intentional

| Count | Rule | Why |
|---|---|---|
| 7 | `js/disabling-certificate-validation` | The opt-in path above. Now scoped per request, still opt-in and default-secure. |
| 1 (critical) | `js/command-line-injection` | `spawn()` with an args array and **no shell**; `cmd` is one of three known values gated by `existsSync`. |
| ~5 | `js/file-access-to-http` | Reading your code and sending it to a configured LLM provider **is the product**. |
| 14 | `js/file-system-race` | Check-then-use on the user's own repo in a local CLI; one site is deliberate (token-checked lock release). |

For individual dismissal with stated reasons once the new baseline lands.

## Note

`mcp-watcher-parity.test.ts` "Scenario 3" still flakes under full-suite concurrency (~1 in 5). Pre-existing — reproduces on pristine `main`, passes in isolation. Unrelated to this PR.

---

# Part 4 — Adversarial review passes

Three passes over this PR's own diff. They found real defects each time, which is why they were worth running.

## Two regressions the TLS scoping introduced

Hand-enumerating which `fetch` calls needed wrapping was wrong **twice**:

- **`doctor.ts`** opted in via `allowInsecureTls` but never wrapped its `/embeddings` probe. Under the old process-global variable it inherited the relaxation; under per-request scoping it did not. Net effect: `openlore doctor` would report a **failed embedding connection to exactly the users who set `skipSslVerify`**.
- **`view.ts`**'s three provider model lookups lost the relaxation they used to inherit from `--insecure`, so `openlore --insecure view` would stop listing models from a self-signed endpoint.

Neither broke a test, because nothing in the suite probes a self-signed endpoint through those paths. Proven end to end against a real self-signed server:

```
withRelaxedTls neutralized  ->  Embedding connection: fetch failed
scoping restored            ->  Embedding connection: 2 dims, 17ms
```

## Made structural instead of remembered

`tls-coverage.test.ts` fails when an https-capable `fetch` is not wrapped, naming file and line. Loopback `http://` calls to OpenLore's own daemon are exempt **with a stated reason**, and a second assertion fails if an exemption stops pointing at a `fetch` — so the list cannot rot into "reviewed and fine" while guarding nothing. Verified against the exact bug above.

One exemption is explicitly **not** an endorsement: `pi/extension.ts` `fetchModels` is recorded as a **pre-existing** gap. The Pi host never calls `allowInsecureTls`, so its provider probe has never honoured `skipSslVerify` — before or after this change. Wrapping it alone would do nothing; making it work needs the Pi host to read config and opt in, which is a behaviour change in its own right.

## Verified nothing else breaks

| Check | Result |
|---|---|
| `init` + `analyze` on a fresh repo | artifacts generate (exercises `file-walker` + `repository-mapper` after `Object.create(null)`) |
| `get_function_body` | `validate` resolves; `foo(.*)\|validate` correctly **not-found** rather than matching the wrong declaration |
| `toDotFormat` with a newline-bearing name | 6 statements, no injected line, balanced quotes |
| `--insecure doctor` / `--insecure --quiet` | exactly one notice / none |
| Public API surface | unchanged — neither new module is reachable through the `exports` map |
| 7 full-suite runs | only intermittent failure is the pre-existing `mcp-watcher-parity` flake, which reproduces on pristine `main` |

### DIFF
--- a/.github/codeql/codeql-config.yml
+++ b/.github/codeql/codeql-config.yml
@@ -8,7 +8,37 @@ name: 'OpenLore CodeQL config'
 queries:
   - uses: security-extended
 
+# ---------------------------------------------------------------------------
+# Scope policy
+#
+# Everything excluded below is excluded for one reason: it is not part of the
+# shipped program, so a dataflow SOURCE inside it describes a scenario that cannot
+# occur in the artifact users run.
+#
+# Findings are NOT suppressed by rule. When a query fires on real shipped code and
+# the behaviour is intended, that alert is dismissed individually in the Security
+# tab with a stated reason — so the rule keeps guarding new code instead of going
+# quiet everywhere.
+# ---------------------------------------------------------------------------
 paths-ignore:
+  # ---- test code -----------------------------------------------------------
+  # Not shipped (the `files` allowlist omits it and `npm run audit:packlist`
+  # enforces that), not reachable by anything but the test runner — and its
+  # fixtures are what made the first baseline unreadable.
+  #
+  # 368 of 458 alerts were IN test files, and the largest rule by a wide margin
+  # (js/insecure-temporary-file, 379 alerts) fires on `writeFile(join(tmpdir(), …))`
+  # in fixtures. That taint also flowed INTO production sinks: 30 further alerts
+  # landed on production `writeFile` calls whose path arrived from a test caller.
+  # gitignore-manager.ts was reported that way despite containing no reference to
+  # `tmpdir()` at all — in production those paths are the user's repository root.
+  #
+  # So this removes a source of noise AND of misattribution. Writing into a
+  # per-test temp directory is the correct thing for a test to do.
+  - '**/*.test.ts'
+  - '**/*.spec.ts'
+
+  # ---- fixtures and vendored code -----------------------------------------
   # Deliberately malformed / adversarial inputs that exist to be parsed by tests.
   # Scanning them reports the fixture, never the code under test.
   - src/core/analyzer/fixtures
--- a/src/api/generate.ts
+++ b/src/api/generate.ts
@@ -6,6 +6,7 @@
  */
 
 import { join } from 'node:path';
+import { allowInsecureTls } from '../core/services/tls-scope.js';
 import { readJsonFile } from '../utils/command-helpers.js';
 import {
   readOpenLoreConfig,
@@ -168,7 +169,7 @@ export async function openloreGenerate(options: GenerateApiOptions = {}): Promis
   // Apply SSL verification setting
   const sslVerify = options.sslVerify ?? openloreConfig.llm?.sslVerify ?? true;
   if (!sslVerify || openloreConfig.generation.skipSslVerify || openloreConfig.embedding?.skipSslVerify) {
-    process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
+    allowInsecureTls('sslVerify=false or config skipSslVerify');
   }
 
   // Create LLM service
--- a/src/cli/commands/doctor.ts
+++ b/src/cli/commands/doctor.ts
@@ -6,7 +6,7 @@
  */
 
 import { Command } from 'commander';
-import { allowInsecureTls } from '../../core/services/tls-scope.js';
+import { allowInsecureTls, withRelaxedTls } from '../../core/services/tls-scope.js';
 import { access, stat, readFile } from 'node:fs/promises';
 import { join, relative } from 'node:path';
 import { execFile } from 'node:child_process';
@@ -496,12 +496,12 @@ async function checkEmbeddingConnection(rootPath: string): Promise<CheckResult |
   try {
     const controller = new AbortController();
     const timeout = setTimeout(() => controller.abort(), DOCTOR_TIMEOUT_MS);
-    const response = await fetch(`${url}/embeddings`, {
+    const response = await withRelaxedTls(() => fetch(`${url}/embeddings`, {
       method: 'POST',
       headers: { 'Content-Type': 'application/json', Authorization: `Bearer ${apiKey}` },
       body: JSON.stringify({ model, input: 'ping' }),
       signal: controller.signal,
-    }).finally(() => clearTimeout(timeout));
+    })).finally(() => clearTimeout(timeout));
 
     const ms = Date.now() - t0;
     if (!response.ok) {
--- a/src/cli/commands/doctor.ts
+++ b/src/cli/commands/doctor.ts
@@ -6,6 +6,7 @@
  */
 
 import { Command } from 'commander';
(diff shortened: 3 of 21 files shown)


clay-good:
### Addressing the 6 CodeQL bot comments (`js/file-access-to-http`)

All six are the **same rule and the same by-design pattern**, and none is a new exposure introduced by this PR.

| File | What the flow actually is |
|---|---|
| `embedding-service.ts:158` | source text read from disk → the configured embedding endpoint |
| `llm-service.ts:1376`, `chat-agent.ts:210` | code/context read from disk → the configured LLM provider |
| `doctor.ts:499`, `view.ts:592`, `view.ts:615` | API key / base URL read from `.openlore/config.json` → the provider being probed |

"Outbound network request depends on file data" is a precise description of **what OpenLore is**: it reads your repository and your provider config, and sends them to the endpoint you configured. Removing the flow would remove the product.

**Why they appeared on this PR:** they are pre-existing alerts that got **re-anchored** to modified lines. Every one of these `fetch` calls was already flagged on `main`; wrapping them in `withRelaxedTls(() => fetch(...))` changed the line, so CodeQL reports them as introduced here. No new data flow was added — the diff wraps existing calls, it does not create new ones.

**Why they don't block:** the rule is `medium`, and the repo's `code_scanning` ruleset gates at `high_or_higher`. `MERGEABLE/CLEAN` reflects that rather than the rule being ignored.

**Disposition:** these belong to the documented "intentional, not fixed" set in the PR description, alongside the TLS opt-out and the `spawn()` finding. They are for **individual dismissal with a stated reason** on `main` once this merges and CodeQL re-baselines — deliberately per-alert rather than by suppressing the rule, so it keeps guarding genuinely new outbound flows in future code.

PR #419, opened 2026-08-30
fix(spec-workflow): stop Generate and Repair presenting unvalidated prose as spec

laurentftech:
## Why

Two failures, both reported by running the Repair skill on a machine where the `openspec` CLI was not working.

**1. An unvalidated spec was indistinguishable from a finished one.** Repair delegated all format correctness to `openspec validate` and said only *"validate strictly"*. With no CLI, nothing validated the edit and nothing said so — the result looked complete and was not valid OpenSpec.

**2. The workflows offered documentation as a specification target.** Reproduced on this repository:

```jsonc
// prepare_spec_generation --domain agents
"receipt": { "state": "complete", "omitted": [], "followUps": [] },
"specOverlap": { "candidateSymbols": 0 },
"domainEvidence": [
  { "path": "LICENSE",            "roles": ["defining", "service"] },
  { "path": "CODE_OF_CONDUCT.md", "roles": ["defining", "service"] },
  …8 markdown/text files + flake.nix
]
```

A `complete` receipt, no omissions, no follow-ups, for a domain of prose — with `LICENSE` labelled a **service**. An agent handed that evidence writes SHALL requirements about a licence, and has nothing to flinch at.

## Approach

OpenLore does not own the OpenSpec format and must not reimplement its rules — the overview spec already delegates the change lifecycle to the CLI. So neither fix validates anything or restates a format. Both **disclose**.

## Validation is disclosed, never performed

Both authoring paths carry it identically — a spec written by Generate is no more self-validating than one edited by Repair:

- `evidence.specValidation` on every page: `{ validatedByOpenLore: false, command, packageResolution, specRoot }`
- the exact command as a terminal-page follow-up
- both skills must report a spec they could not validate as **NOT validated**, never as complete

The command is `openspec validate --specs --strict`. Not `--all`: an unrelated invalid change in flight would fail a perfectly valid baseline spec. Not the default mode: it accepts warnings the skills promise to treat as failures.

`packageResolution: "unresolved"` is a disclosed unknown, not "the CLI is absent" — a global install resolves from neither scope — so the follow-up is emitted either way.

## Neither skill restates the format

The prior text had already drifted: it said `GIVEN/WHEN/THEN` where OpenSpec's own template says `WHEN/THEN`.

Each skill now takes its shape from an existing baseline spec — the one Repair edits, a sibling for Generate — and lets `openspec validate` judge it. Both explicitly do **not** follow `openspec instructions specs --change <id>`: that describes the change-local delta form (`## ADDED Requirements`, …), and writing its operation headers into a baseline spec corrupts what archive later merges.

## Documentation never defines a domain

Prose, licences, and project meta are classified `supporting`, so a documentation-only tree is dropped by the **existing** `non-defining-only` rule — no new mechanism, no new tuning constant. `supporting` rather than `excluded`, so a code domain keeps its own docs as footprint evidence.

Ambiguity resolves toward **under-matching**, the safe direction: a missed prose file is merely supporting, whereas a misread source file would silently remove its domain.

- project stems match only in their conventional upper-case spelling — `bin/readme`, `scripts/changelog`, `license.ts` stay source
- `LICENSE-MIT`, `LICENSE_APACHE`, `COPYING.LESSER` are recognised
- `.txt` excepts the build-config names sharing it (`requirements*.txt`, `constraints*.txt`, `CMakeLists.txt`), anchored on the whole basename so `docs/product-requirements.txt` stays prose
- checked before the config flag, so `docs/config.md` stays prose

## A behavior-free domain is disclosed, not decided

`evidence.domainBehavior` reports whether anything in the domain could carry a requirement; the terminal page carries a stop-and-ask follow-up when it cannot.

The trigger is **positive evidence of prose** — every defining file is documentation — never the absence of extracted signatures, which a declarative Vue/Svelte domain or an `app.get(...)` bootstrap also lacks.

- symbols are counted, not the files carrying them (a 25-symbol file was reporting `1`)
- only *defining* symbols count: a requirement anchors to the implementation, never to a test
- with no analyzed domain the counts stay `null` — never a fabricated zero — and the stop fires only when every cited source is prose, so `overview` and `architecture`, which own no source by design, stay repairable
- an anchored citation (`README.md#Installation`) is classified by the file it names, not the fragment

## Also fixed: a pre-existing spec-root bug

Found while verifying the above. `handleGetSpec` hardcoded `openspec/`, so a repository with a configured `openspecPath` got `spec-not-found` for specs plainly on disk — Repair could not run at all there. Its reported path and provenance now name the file actually read; querying git about an unrelated clean path could otherwise serve a modified spec as `reviewed-corpus`. The prose-evidence corpus filter uses the configured root too, normalized so a `./openspec` value matches.

## Verification

- **9010 passing.** The 4 remaining failures are pre-existing and reproduce on a clean tree (three assert an OpenSpec package version, one is git-hooks drift)
- `openspec validate --all --strict` → 104/104
- `tsc --noEmit` clean, `eslint` clean
- Spec deltas in the open `harden-spec-workflow-lifecycle` change: `analyzer` and `mcp-handlers` requirements, `generator` requirement extended, tasks 7.6–7.8
- New tests cover: both paths disclosing validation, the relocated spec root, prose-only domains, a corpus-level spec staying repairable, test-only symbols, anchored citations, licence variants, extensionless executables
- Reviewed by Codex across many rounds; every finding either fixed or verified as a false positive

https://claude.ai/code/session_01YGDj2YnUQdVXARyo1LSrr1


clay-good:
Hardened and rebased onto current main. The added commit routes configured spec roots through OpenLore's existing repository-confined resolver, prevents MCP spec reads outside the project, supports a spec corpus rooted at `.`, and broadens conventional project-document classification without misclassifying lowercase executables. Added regression coverage for each boundary. Local verification: strict OpenSpec validation (101/101), focused #419 tests (116/116), equivalence lane (355/355), remaining unit lane (8,686 passed; 5 contention timeouts all passed in isolated one-worker rerun, 90/90), deterministic integration (35/35), build, lint, typecheck, dependency audit, package-content audit, and diff check.

PR #449, opened 2026-09-05
fix(install): wire OpenLore's own CLI, not the npx shim that opens a console window

laurentftech:
> Follow-up to #431, from live use on Windows 11 **after** it merged. #431 fixed the console
> flashes OpenLore itself caused. This fixes the one it could not reach: the flash is not
> OpenLore's spawn at all — it is `npx`'s, and the process doing it is the host agent.
>
> **Rebased onto #448.** That PR closed the same bug class from the inside — every subprocess
> in `src/` now sets `windowsHide`, enforced by an import-aware guard. The two are disjoint:
> #448 governs spawns OpenLore performs, this one governs the command string OpenLore *writes
> into a host config* for someone else to spawn. No file overlaps; the rebase was clean, and
> `windows-hidden-spawn-guard.test.ts` passes here.

# The finding

`openlore install` wires every host command as `npx --yes openlore <args>`.
`resolvePlatformCommand` already removes the Windows `.cmd` boundary for **`npx` itself** —
that part works. But npx then launches the **target package's** bin through *its* shim.

Captured live with a process monitor while reproducing the user's report:

```
cmd.exe 16292   C:\WINDOWS\system32\cmd.exe /d /s /c openlore orient --inject
cmd.exe  2172   C:\WINDOWS\system32\cmd.exe /d /s /c openlore mcp --preset substrate
```

Each is a real console window, each with its own `conhost.exe`. The `UserPromptSubmit` hook
fires on **every agent turn**, so this is one window per turn. The reporting user saw it live,
mid-session, twice, while this was being diagnosed.

`%APPDATA%\npm\openlore.cmd` is what npx resolves to; a `.cmd` can only be launched through
`cmd.exe`. Nothing in #431 could have caught it: the guard tests there cover `detached: true`
spawns and `git` spawns **in OpenLore's own source**, and this spawn is in neither — it is not
our code.

# The fix

`resolveOpenloreCommand` in `src/utils/platform-command.ts`, the module that already owns
"avoid the Windows shim boundary".

`openlore install` **is** openlore, so the CLI entry beside the running module is the same
build the user just invoked. Wire that with the absolute Node executable:

```
"command": "C:\Program Files\nodejs\node.exe",
"args": ["…\openlore\dist\cli\index.js", "mcp", "--preset", "substrate"]
```

No shim, no `cmd.exe`, and one process hop fewer on every hook fire.

Migrated the three adapters that wire openlore — `claude-code`, `continue`, `cursor`. No
adapter constructs the `npx` form any more.

## It falls back, never breaks

Three cases keep the previous `npx --yes openlore` form:

| Case | Why |
|---|---|
| Running from TypeScript source | No built sibling next to the module |
| Entry inside an `_npx` cache | npm evicts that path; it must never reach a config file |
| Windows node path not absolute | The existing loud refusal is preserved, not softened |

## Trade-off, deliberate

The wired path is absolute, so it binds the config to this install location. A global npm
install keeps that path across upgrades; a moved install needs `openlore install` re-run, where
`npx` would have re-fetched. That is the price of not opening a window on every turn.

# What is NOT in here: `--watch`

Worth recording, because it was the first hypothesis and it was **wrong**.

The index looked stale (pinned to a pre-rebase commit), and `openlore mcp --preset substrate`
carries neither `--watch` nor `--daemon`, which reads like "no continuous watching is possible".
It is not: `--watch-auto` **defaults to true** (`mcp.ts:3170`) and starts an in-process
`McpWatcher` with `selfRebuild: true` on the first tool call carrying a directory
(`mcp.ts:2830`).

Verified empirically rather than argued: a new file with two unique symbols was created with
**no daemon running and no MCP tool call afterwards**. The index republished **2 seconds later**
— 497→498 files, 4564→4566 functions. The watcher is fine and needed no change.

# Verification

- `platform-command.test.ts`: **16/16**, 5 new — the no-`cmd.exe` invariant asserted directly
  (`expect(line).not.toContain('.cmd')`), plus all three fallbacks and the non-Windows
  pass-through.
- Installer suites: **18 failed / 105 passed**, against an **18 failed / 100 passed** baseline
  measured by checking the touched files out of `origin/main` (post-#448) and re-running. Same
  18 pre-existing Windows-local failures (file modes, symlinks); the +5 are the new tests.
  **Zero regressions.**
- #448's new `windows-hidden-spawn-guard.test.ts` passes unchanged on this branch (33/33 with
  `git-exec` and `platform-command`).
- `tsc --noEmit` and `eslint` clean.
- Confirmed on the reporting machine: re-running the monitor, the old form spawns a `cmd.exe`,
  the new form spawns none.

---

# Second half: the watcher had never run on Windows

The console-flash fix above came from asking why a window opened. The same session asked the
next question — is the index really kept fresh? — and that exposed a coverage hole rather than a
behaviour bug.

**The watcher works.** Measured on Windows 11 with no daemon running and no MCP tool call in
between: a new file was indexed in **1s** (497→498 files, 4569→4571 functions), a modification in
**1s** (+1 function, **+2 call edges** — it parses content, it does not count files), and a
deletion in **2s**, restoring the baseline to the exact number. `--watch-auto` defaults to true
and arms an in-process `McpWatcher` on the first tool call carrying a directory.

## But almost none of that ran in CI

- `mcp-watcher.integration.test.ts` — the real-chokidar file, including the
  delete-a-node-from-the-graph coverage — was in **no CI job at all**. It is absent from
  `test:integration:ci`.
- `mcp-watcher-incremental.test.ts` and `mcp-watcher-parity.test.ts` are excluded from
  `npm run test:unit`.
- vitest is ubuntu-only, so **the watcher had never executed on Windows** — the platform where
  path separators and file locking differ most.

## Running it on Windows found 9 real failures

**Seven were one defect**: `mcp-watcher.test.ts` built its fixture root from the POSIX literal
`'/tmp/proj'`, which is not an absolute path on Windows, so `safeJoin`'s lexical containment check
could never hold and *every* watcher construction threw `Path traversal blocked`. Same class as
513b12dc. A platform-native `FIXTURE_ROOT` makes both platforms assert the same behaviour →
**35/35**.

**Two are scenarios Windows cannot set up**, now `skipIf(win32)` with the reason written down:
`chmod(0o000)` does not express "exists but unreadable" there (Node maps the mode to the read-only
attribute alone), and `symlink` needs elevated privileges.

## The ninth is NOT fixed, and now says so

`until()` returned **silently** on timeout, so a flush that never landed surfaced as a baffling
content assertion. It now throws:

    until(): condition was still false after 20000ms (the flush never landed)

That is a diagnosis, not a fix. 20s for a 20ms-debounced flush is not a budget problem — the flush
genuinely never happens. Characterised: passes **8/8 in isolation**, fails **~1 run in 3** alongside
other files, and `--no-file-parallelism` does **not** help, so it is cross-file state, not load.
Two of my hypotheses were wrong along the way (a double read in the test; then parallel contention);
both are recorded in the commit so the next investigation does not repeat them. **Tracked in #451**,
with the full reproduction, the disproved hypotheses, and the suggested next step.

# A real branch switch is now tested

The watcher rebuilds the graph when `.git/HEAD` moves. Every existing test **injected** that
trigger — `mcp-watcher.graph-rebuild.test.ts` drives `scheduleGraphRebuild` directly, proving the
coalescing window and nothing about how the trigger is reached. The untested half is the chain:

```
git checkout  ->  .git/HEAD rewritten  ->  chokidar ref watch  ->  'head-change'
```

Two tests with real git, real chokidar, real filesystem — nothing stubbed. A `git checkout -b` must
reach the rebuild lane; a bare `git add` must **not** (`.git/index` moves on every stage, and
treating that as a branch switch would rebuild the graph on routine work). Mutation-checked:
disabling the HEAD basename check turns the first red and leaves the second green.

# CI now runs the watcher on Windows

`test:integration:ci` gains the real-fs file, and Windows Smoke gains two steps — the watcher unit
files, and the real-git integration file. `mcp-watcher-incremental.test.ts` is excluded **on
purpose, with the reason in the workflow**: wiring a known flake into CI buys noise, not signal.

Verified locally on Windows 11, invoked exactly as the workflow does: unit step **100 passed / 2
skipped**, integration step **13/13 twice**. Both steps pass the PowerShell parser; `ci.yml` parses.

# Corrections to earlier claims

Three, recorded because each was stated confidently before being checked.

**`preflight/index.ts` and `heap-sizing.ts` as unguarded spawn sites** — wrong. #448's guard names them explicitly as **deliberate exemptions**: they use
`stdio: 'inherit'`, so the child attaches to the console the parent already has — no new window
is created, and forcing `CREATE_NO_WINDOW` would suppress interactive output that must reach the
real console. Nothing is owed there.

**"The watcher cannot run continuously in this installation"** — wrong. `--watch-auto` defaults to
true; the watcher was running the whole time, and a file was indexed 2s after creation with no
daemon and no tool call. The `--watch` flag I first proposed adding to the installer was dropped
from this PR as unnecessary.

**"The delete-from-graph test is covered by CI"** — wrong. Its file runs in no CI job; that is what
this PR fixes.

# What is not proven

- The intermittent flush failure is characterised, not fixed, and its test file is excluded from
  the new Windows step. Windows watcher coverage is therefore real but not complete.
- The Windows steps are verified on one Windows 11 machine and have not yet run on a GitHub runner.
- Only the three openlore-wiring adapters were migrated. If another host adapter is added later
  and reaches for `npx` directly, nothing structurally stops it — there is no guard test for
  that shape yet, and I did not add one.
- Verified on one Windows 11 machine, like #431.
- CI has not run on this branch yet at the time of writing.


clay-good:
## Hardening pass — 3 defects found and fixed, then merged-ready

Reviewed the change and dogfooded it as a **real built install**, which is where this
change actually lives. That turned out to matter: **every install test runs from
TypeScript source**, where no built CLI sits beside the module, so all of them exercised
the `npx` **fallback** — none saw the shape a user gets. Three defects were hiding in that
gap.

### 1. Uninstall stranded the entry it had just written (regression, blocking)

`isOurMcpEntry` identifies a marker-less `mcpServers.openlore` by the literal `openlore`
argument in its argv. The direct form drops that argument, so the marker-loss fallback
added in #455 matched **nothing** — the exact failure that fallback exists to prevent,
in **every** repository.

Reproduced against a real install: wire, delete `_openlore` (what any tool that rewrites
`~/.claude.json` and drops unknown keys does), uninstall → the entry survived. Now fixed
by also recognising the CLI-entry path shape, kept beside the emitter so the two cannot
drift. **Mutation-checked**: restoring the old predicate turns the new test red.

The first version of that shape check was itself wrong — it required a path segment
`=== 'openlore'`, which a development worktree does not have, and the real uninstall
still removed nothing. Caught by re-running the real uninstall, not by the suite.

### 2. The hook command string was not shell-safe on POSIX

`formatPlatformCommand` produces a **string a host runs through a shell**, and since this
change it carries absolute paths — including the user's home directory — on **every**
platform, not only Windows. Double quotes leave `$` and backticks live, so a home
directory containing either turned the `UserPromptSubmit` hook into a substitution.
POSIX formatting is now single-quoted; the cmd.exe form is unchanged.

`platform` is now a **required** argument rather than defaulting to `process.platform`.
That default was already silently wrong in `openlore update`, which formats a
Windows-resolved invocation — the required argument surfaced it immediately.

### 3. The absolute path had no failure story — `doctor` now tells it

The pinned path is this PR's deliberate trade-off, and it is the right one. But a moved
install, an uninstalled global, or a removed Node version previously showed up **only** as
a hook that fails silently on every turn. `openlore doctor` now names the missing launcher
and offers the re-wire remediation it already has:

```
warn  MCP wiring  .mcp.json points at a launcher that no longer exists: /gone/openlore/dist/cli/index.js
      Run 'openlore install --agent claude-code --force' to re-wire it to this install
```

Verified live, both directions.

### The guard test the PR said was missing

Added: the three adapters are applied with an injected CLI entry, and nothing they write
to `.mcp.json`, `.claude/settings.json`, `.cursor/mcp.json`, or `.continue/config.json`
may contain `npx` or `.cmd` — on `win32` and on `linux`.

### Verification

- `tsc --noEmit` and `eslint` clean.
- Install + platform-command + doctor + update suites: **266 passed**.
- Full unit suite: the 15 failures are **pre-existing on this branch** — confirmed by
  stashing the hardening and re-running the same files (identical set; they are local
  git-hook/subprocess/npm-audit environment failures, and CI is green on all of them).
- End-to-end on a real built install: wire → the hook command executes → uninstall removes
  the entry after the marker is gone and leaves the user's own `numStartups` intact.

**Verdict: LGTM with the above applied.** The `--watch` correction and the watcher's
Windows CI coverage are solid, and the "what is not proven" section is honest about the
flush flake (now closed separately in #453).

Issue #452, opened 2026-09-05
Windows unit-test backlog: the files on the tracked deny-list


clay-good:
Follow-up to #434, which added the **Windows Unit Tests** job. That job runs the whole unit
suite on `windows-latest`: **9,284 of 9,538 tests pass across 493 files** (before it, exactly
one file ran on Windows).

The files that do not yet pass are tracked in
[`.github/windows-unit-exclusions.json`](../blob/main/.github/windows-unit-exclusions.json),
with a reason on each entry.

The list is enforced in both directions by `scripts/windows-unit-report.mjs`, so it can only
shrink: a listed file that starts passing fails the build until its entry is deleted, and any
unlisted failure is a new regression. Nothing here can rot silently.

**To pick up a slice:** download the `windows-unit-report` artifact from any run of that job —
`gh run download <run-id> -n windows-unit-report` — and read the actual failure messages. Treat
each entry's `reason` as a hypothesis; #458 found two of them describing the wrong defect.

clay-good:
Adding one item found while building the gate: `src/core/drift/git-diff-corpus-materialization.test.ts` is **flaky on Windows, not deterministically failing**.

Observed across four runs of the same tree: ENOENT in the initial probe, then pass → fail (`rejects invalid UTF-8 corpus bytes`) → pass in three consecutive CI runs, with `--retry=2` already in effect.

It reads a `git cat-file --batch` child's stdout, so a stream race is the likely cause, but it is **not root-caused** — that needs a Windows machine to debug rather than a 12-minute CI cycle.

It is marked `flaky: true` in the deny-list, which exempts it from both enforcement rules (a file that oscillates would otherwise fail the build whichever list it sat on) and reports it on every run so it cannot become invisible. That flag is a bug to fix, not a state to keep — it is the one entry that should be looked at before the deterministic ones.

clay-good:
### First slice in flight: the "suspected production bugs" — #458

Diagnosed from the `windows-unit-report` artifact of a green `main` run rather than by
inspection, so each is quoted by its exact failing assertion. **Four of the five were as
described; the other two labels were wrong.**

| File | Verdict |
|---|---|
| `spec-mapper.test.ts` | real bug — `adrPath` and spec keys carried `\` |
| `syncer.test.ts` | real bug — `modifiedSpecs` carried `\` |
| `test-generator.test.ts` | real bug — `outputPath` carried `\` |
[...]

PR #458, opened 2026-09-06
fix(paths): a repository-relative path OpenLore serves is POSIX on every platform

clay-good:
**Status:** LGTM — closes the *suspected production bug* slice of #452 (4 of 5 files, 10 of 11 tests).

## What was wrong

`path.relative()` and `path.join()` yield backslashes on Windows. Three places put that
result into a value OpenLore **serves or persists** rather than into a filesystem call, so
the same repository produced different answers depending on which platform analyzed it.

| Site | What Windows emitted | Why it matters |
|---|---|---|
| `spec-mapper` | `openspec\decisions\adr-0003-caching.md`, and spec keys with `\` | a mapping built on Windows does not match the POSIX spec corpus for the same repository |
| `syncer` | `modifiedSpecs` as `openspec\decisions\adr-….md` | no consumer of the POSIX corpus matches it |
| `test-generator` | `outputPath` as `spec-tests\auth\user-login.spec.ts` | a generated plan differs between a Windows and a POSIX run |

## How it was fixed

All three go through `toRepositoryPath` — the rule the walker already enforces for its own
keys. **Nothing about the write target changes**: Node accepts `/` on Windows, and every one
of these values is re-joined against the root before use.

`architecture-writer` was **not** a production bug and the backlog labelled it wrongly. It
returns a real filesystem path, where the native separator is the correct output; the test
hard-coded the POSIX form. The expectation is now built with `join()`.

## Replication / proof

Diagnosis came from the `windows-unit-report` artifact of a green `main` run — not from
guessing. Each failing assertion, quoted from that report:

```
syncer.test.ts     expected 'openspec\decisions\adr-0001-use-redis…' to match /^openspec\/decisions\/adr-/
spec-mapper.test   expected 'openspec\decisions\adr-0003-caching.md' to contain 'openspec/decisions/adr-0003-caching.md'
test-generator     expected 'spec-tests\Auth\UserLoginTest.java' to be 'spec-tests/Auth/UserLoginTest.java'
architecture-write expected '\my\project\.openlore\analysis\ARCHIT…' to be '/my/project/.openlore/analysis/ARCHIT…'
```

The four entries are deleted from `.github/windows-unit-exclusions.json`. **That list is the
proof**: it is enforced in both directions, so these files failing again fails the build, and
a listed file that quietly started passing fails it too. POSIX suites are unchanged — 3,940
tests passing across the touched trees, `tsc` and `eslint` clean.

## Notes

The fifth file, `index-bundle`, is a **different defect than the list claimed**. The same
report shows its imported answer reporting a different `confidenceBoundary.basis.directEdges`
count than the identical query against the local index — an edge-resolution divergence in the
`.olbundle` round trip, not a separator inside a string. It stays on the list with the reason
corrected, because fixing it needs a Windows machine rather than a 12-minute CI cycle.

### DIFF
--- a/.github/windows-unit-exclusions.json
+++ b/.github/windows-unit-exclusions.json
@@ -204,6 +204,11 @@
       "reason": "POSIX-specific fixture or assertion"
     },
     {
+      "file": "src/core/decisions/syncer.test.ts",
+      "failingTests": 1,
+      "reason": "'round-trips a decision constraint through the owning spec and ADR before purging': loadDecisionConstraintState reports one malformedFinding on Windows where POSIX reports none. NOT the separator defect the other four tests in this file had (fixed in the POSIX-repository-path change) \u2014 a durable-projection/marker read that behaves differently there, and the report artifact truncates the finding, so it needs a Windows machine to name."
+    },
+    {
       "file": "src/core/drift/git-diff-corpus-materialization.test.ts",
       "failingTests": 1,
       "reason": "FLAKY on Windows, not deterministically failing: observed passing, then failing on 'rejects invalid UTF-8 corpus bytes', then passing across three consecutive CI runs of the same tree, and failing with ENOENT in an earlier probe. Reads a `git cat-file --batch` child's stdout, so a stream race is the likely cause. Not root-caused: needs a Windows machine to debug, not a 12-minute CI cycle.",
--- a/.github/windows-unit-exclusions.json
+++ b/.github/windows-unit-exclusions.json
@@ -163,12 +163,6 @@
       "reason": "mocked-fs unit test built on POSIX-absolute fixture paths"
     },
     {
-      "file": "src/core/analyzer/architecture-writer.test.ts",
-      "failingTests": 1,
-      "reason": "architecture artifact records its own path with platform separators",
-      "suspectedProductionBug": true
-    },
-    {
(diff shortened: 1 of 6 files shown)


clay-good:
## Windows verdict: green

```
windows-unit-report: 9284/9538 tests passed across 493 files;
                     48 of 48 deny-listed files still failing, 1 tolerated flaky, 0 new.
windows-unit-report: OK — no new failures, and nothing stale on the list.
```

`0 new` is the assertion that matters: `architecture-writer` and `test-generator` came back
clean and stay off the list, and `spec-mapper` now passes with the POSIX contract asserted
directly. Every other job is green as well.

### Correction to this PR's original claim

It closes **9 of 11 tests, not 10.** `syncer` is back on the list with **one** test rather
than five. Four were the separator defect and are fixed; the fifth is unrelated —
`loadDecisionConstraintState` reports one `malformedFinding` on Windows where POSIX reports
none, and the report artifact truncates the finding, so naming it needs a Windows machine.
Recorded on the entry rather than left implied.

Also worth stating plainly: the first push of this PR **failed** the Windows job, and that
was the point of pushing it. `spec-mapper.test.ts` was asserting its expected `specPath` with
`join()` — the native separator, i.e. the behaviour this change replaces — which is invisible
on POSIX, where `join()` and the POSIX form are the same string. The deny-list gate caught it.

PR #460, opened 2026-09-06
fix(test): stop the durability suite manufacturing the contention it trips over

### LINKED ISSUE #457: Windows: watcher loses a change when the atomic rename hits EPERM
On Windows, the watcher can lose a file change when publishing the index. The atomic commit
rename fails with `EPERM` after exhausting its retry ladder, and the watcher gives up:

```
[mcp-watcher] error: EPERM: operation not permitted, rename
  '…\.openlore\analysis\.llm-context.json.tmp-5368-5' -> '…\.openlore\analysis\llm-context.json'
[mcp-watcher] gave up on 1 change(s) after 3 attempt(s) — run analyze to reconcile: foo.ts
```

**Impact:** the change is reported, not silently dropped, but the index stays stale for that
file until a full `openlore analyze`.

## Reproduce

Windows 11, Node 22+. Fails about 1 run in 4–6:

```
node node_modules/vitest/vitest.mjs run src/core/services/mcp-watcher-flush-durability.test.ts
```

## What we know

`atomicWriteFile` already retries this exact error set — `RENAME_CONTENTION_DELAYS_MS` in
`src/core/decisions/atomic-store.ts` is eight attempts over ~3.2s. So the ladder is being
exhausted, not bypassed. The fix depends on who holds the destination file:

- an external process (Defender, Search Indexer) → the ladder is too short
- one of our own readers → the read side needs `FILE_SHARE_DELETE`; no ladder length fixes it

Ruled out: not 8.3/short-path related, not introduced by #449, and not the test's 20s budget
(the watcher has already given up before it expires).

laurentftech:
Follow-up to #457, from a Windows box — which is what that issue was waiting for.

`mcp-watcher-flush-durability.test.ts` failed about **1 run in 4** on Windows, always the same
way: the watcher's artifact publish lost its rename to `EPERM`, gave up after 3 attempts, and the
assertion failed for a reason unrelated to flush durability. The file is **not** on the Windows
deny-list, so it was queued to redden a `windows-unit` run.

## Evidence

Measured with a Restart Manager probe fired at the moment the retry ladder in `atomicWriteFile`
exhausts (no Sysinternals on the machine; the probe was validated against a known lock first).
Three findings, in the order they changed the conclusion:

1. **The holder is our own process** — the same pid that wrote the temp file, readable straight
   off `…llm-context.json.tmp-<pid>-<n>`. Not Defender, not the Search Indexer.
2. **Every open shape blocks the replace**, `O_RDONLY` included — so `readFileConfined`'s
   `O_NOFOLLOW` is not special:
   ```
   O_RDONLY                                   -> rename FAILED EPERM
   O_RDONLY | O_NOFOLLOW  (readFileConfined)  -> rename FAILED EPERM
   O_RDWR                                     -> rename FAILED EPERM
   no handle at all                           -> rename SUCCEEDED
   ```
3. **`FILE_SHARE_DELETE` does not rescue it.** PowerShell holds the destination under an explicit
   share mode while **Node** performs the rename, so the call under test is the real
   libuv `MoveFileExW(MOVEFILE_REPLACE_EXISTING)` rather than a .NET approximation:
   ```
   reader WITHOUT share-delete -> node rename FAILED EPERM
   reader WITH    share-delete -> node rename FAILED EPERM
   ```

(3) refutes the premise recorded in #457 — that Node's share flags mean our own reads cannot
block the rename — and collapses that issue's fork. There is no share mode a reader can adopt, so
**retrying is the architecture, not a workaround**.

## The change

**`mcp-watcher-flush-durability.test.ts`** — `until()` polled `readFile(contextPath)` every 10ms.
The watcher publishes that exact file by temp+fsync+rename, so the harness held a descriptor
across most of the publish window and manufactured the contention it then tripped over. The new
`untilContext()` polls every 100ms and reads only when the mtime has moved: more than an order of
magnitude less duty cycle, same assertion — the content still has to land.

**`atomic-store.ts`** — comment only. `RENAME_CONTENTION_DELAYS_MS` attributed the hold to "an
antivirus or Search Indexer", with our own reader as an afterthought. The evidence inverts that,
and the earlier text sent this investigation looking for Defender. Corrected, with the
share-delete result recorded so the next reader does not have to re-derive it.

## Verification

- **0 failures in 15 runs**, from a ~1-in-4 baseline.
- Clean under a short (8.3) `TEMP` too — the shape a Windows runner hands out.
- The 4 failing files under `src/core/decisions` are pre-existing Windows deny-list entries;
  `atomic-store.test.ts` fails the same 2 tests with this change stashed.

## What this deliberately does NOT do

**It does not close #457.** The product behaviour is real: a concurrent reader can still cost a
publish, and on a long-lived server that reader is a tool call serving `llm-context.json` while the
watcher writes it. Production read pressure is far below this harness's — which is why the 1-in-4
rate was never representative — but the mechanism stands, and the issue keeps it.

### DIFF
--- a/src/core/decisions/atomic-store.ts
+++ b/src/core/decisions/atomic-store.ts
@@ -43,12 +43,25 @@ let tmpCounter = 0;
  * Backoff for a rename that lost a race with another handle on the destination.
  *
  * POSIX `rename(2)` replaces the destination unconditionally. Windows does not:
- * `MoveFileExW(REPLACE_EXISTING)` needs DELETE access to the target, so it returns
- * `EPERM`/`EACCES`/`EBUSY` while ANY other handle is open on it without
- * `FILE_SHARE_DELETE` — an antivirus or Search Indexer scanning the file it just
- * saw written, or a reader that opened it a millisecond earlier. The window is
- * milliseconds wide and the operation is idempotent (the temp file is still
- * there, fully fsync'd), so the correct response is to wait and try again.
+ * `MoveFileExW(REPLACE_EXISTING)` returns `EPERM`/`EACCES`/`EBUSY` while ANY other
+ * handle is open on the target. The window is milliseconds wide and the operation is
+ * idempotent (the temp file is still there, fully fsync'd), so the correct response
+ * is to wait and try again.
+ *
+ * WHO holds it is usually US, not an antivirus — measured on Windows 11 / Node 26
+ * while investigating issue #457, after an earlier version of this note guessed at a
+ * scanner and sent the investigation the wrong way:
+ *
+ *   - every open shape blocks it, `O_RDONLY` included (so `readFileConfined`'s
+ *     `O_NOFOLLOW` is not special), and
+ *   - `FILE_SHARE_DELETE` does NOT rescue it. A reader that explicitly grants
+ *     share-delete still blocks the replace, so there is no share mode a reader can
+ *     adopt to make this go away.
+ *
+ * That second point is why this ladder is the ARCHITECTURE and not a workaround: with
+ * nothing to fix on the read side, waiting for our own reader to close its descriptor
+ * is the only way through. Any concurrent read of the artifact — a tool call serving
+ * `llm-context.json` while the watcher publishes it — can cost a publish this way.
  *
  * Observed, not theorised: a Windows CI runner produced
  * `EPERM: operation not permitted, rename '….llm-context.json.tmp-…' -> '…llm-context.json'`
--- a/src/core/decisions/atomic-store.ts
+++ b/src/core/decisions/atomic-store.ts
@@ -58,6 +58,23 @@ let tmpCounter = 0;
  *     share-delete still blocks the replace, so there is no share mode a reader can
  *     adopt to make this go away.
  *
+ * The second point is the load-bearing one, so here is how to re-measure it rather than
+ * take it on trust. Hold the destination open with `CreateFileW` called DIRECTLY — the
+ * share flags then are exactly what Win32 receives — while `node` performs the rename, so
+ * the call under test is the real libuv `MoveFileExW(MOVEFILE_REPLACE_EXISTING)`:
+ *
+ * ```
+ * share READ|WRITE (no delete)  -> node rename FAILED EPERM
+ * share READ|WRITE|DELETE       -> node rename FAILED EPERM
+ * share DELETE only             -> node rename FAILED EPERM
+ * no handle at all              -> node rename SUCCEEDED
+ * ```
+ *
+ * Both halves of that harness matter. A first attempt used .NET's `File.Move(overwrite)`
+ * to do the replacing and reported `ACCESS_DENIED` for every arm including the control —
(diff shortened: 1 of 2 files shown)


clay-good:
Reconciled onto `main` — and I need to retract something of mine that this PR refutes.

**#459 landed underneath this branch**, and the merge left `atomic-store.ts` contradicting
itself: your corrected note on the shared ladder says the holder is usually our own reader,
while the note I added on the Windows extension still claimed it "cannot be one of ours" and
named Defender. I have rewritten my block; yours is untouched.

**My reasoning was wrong, and your evidence is better than mine.** I inferred the holder's
identity from a *clean negative* — a `windows-latest` runner not hitting the contention in 10
runs while the reporting machine failed ~1 in 5 — and concluded the holder had to be external.
That inference does not hold: the contention is timing-sensitive, so an idle runner simply
never lands in the window. Not reproducing says nothing about who holds the handle. Reading the
holding pid straight off the temp file name does, and it says us.

Your point (3) also refutes a premise I had stated elsewhere in this investigation — that
libuv's `FILE_SHARE_DELETE` means our own reads cannot block the rename. I had flagged it as an
argument rather than evidence; testing it with PowerShell holding the destination while *Node*
performs the rename is the right way to settle it, and it came back the other way.

The ladder extension stands, and your evidence argues for it more strongly than my original
reasoning did: if no share mode lets a reader step aside, outwaiting that reader is the only way
through, so the ladder's length is the entire mitigation. I have rewritten the comment to say
that, and to record the retraction rather than quietly delete it.

Also acting on your "what this does NOT do": I had closed #457 on my refuted conclusion.
**Reopened.** The product behaviour you describe — a tool call serving `llm-context.json` while
the watcher publishes it — is real and unfixed; the retry window is mitigation, not a fix.

`tsc`, `eslint`, and both touched suites pass locally (22 tests). CI next.

laurentftech:
Took your merge as-is — the two blocks agree now, and recording the retraction in place rather than deleting it is the better call.

Verified on the reporting machine after your merge: `tsc` clean, and the durability suite is **0 failures in 10 runs** (it was ~1 in 4 before the polling change). So the reproducer machine agrees with your local result.

Added one comment-only commit, for the claim that carries the most weight — that `FILE_SHARE_DELETE` does not rescue the rename. If a reader could step aside, the ladder length would not be the whole mitigation, so the note now carries the *procedure* rather than only the conclusion: hold the destination with `CreateFileW` called directly, so the share flags are exactly what Win32 receives, while `node` performs the rename.

It also records the harness that got it wrong first. My initial attempt used .NET `File.Move(overwrite)` as the replacing call and returned `ACCESS_DENIED` for every arm **including the control** — not the same call, and read carelessly it would have "confirmed" the right conclusion for the wrong reason. Given this thread has twice lost time to an inference presented as a measurement, a wrong instrument that happens to agree seemed worth naming too.

Agreed on reopening #457: the retry window is mitigation, not a fix, and a tool call serving `llm-context.json` while the watcher publishes it is still a real staleness window.

PR #466, opened 2026-09-06
fix(serving): an unservable index says WHY, instead of reporting absence

laurentftech:
## What we observed

Every MCP tool call on a real repository answered:

```
No analysis found. Run analyze_codebase first.
```

The analysis was there — 234 MB of it. What had actually happened:

```
generation.json          published 23:33
llm-context.json         written   23:36

dependency-graph.json    MISMATCH
fingerprint.json         MISMATCH   (447 -> 449 bytes)
llm-context.json         MISMATCH
refactor-priorities.json MISMATCH
repo-structure.json      ok
```

Four of five artifacts no longer matched their published generation. No writer was running, and the state had been settled for five minutes — so this was not a mid-write interval. The integrity gate did exactly its job and refused to serve unverified bytes.

**Then it reported that refusal as an absent index.**

## Why the wording is the bug

`readCachedContext` already **distinguishes** these cases. It emits a different telemetry `reason` for each —

```ts
emit(..., { reason: 'artifact_not_a_regular_file' });     return null;
emit(..., { reason: 'analysis_generation_unavailable' }); return null;
emit(..., { reason: 'analysis_generation_changed' });     return null;
```

— and then returns a bare `null`. The diagnosis is computed, recorded, and discarded at the `return`. Every caller collapses three situations into one sentence that is correct for exactly one of them.

For the others it reports a **failed integrity check as a missing index**, which is precisely the quiet downgrade that `loadPartialFirstRun`'s own docstring says this lane exists to prevent:

> A published artifact that failed an integrity gate above **must keep failing**: masking a corrupt or mid-rewrite index with a partial one would turn a loud problem into a quiet downgrade, which is the opposite of what this lane is for.

The policy is right. The message is that downgrade, expressed in words.

**And it hides the incident.** A user reads "no analysis", runs `analyze`, it works, and the lost publish is never reported. That is nearly what happened here: seeing it at all took comparing the manifest's hashes by hand.

Worth noting what the codebase already says about this state — the comment guarding it describes a brief window:

> A writer updates artifacts in place and publishes the manifest last. **During that interval** the old generation id is still visible…

Ours lasted five minutes with nothing writing. The guard behaved correctly; the publish never completed. Whether that is a lost publish under rename contention (#457) or a flush path that does not republish is **not** established here, and this PR does not claim it — it makes the state say its own name so the next occurrence is reportable instead of invisible.

## What this adds

`diagnoseIndexUnservable(directory)` names the reason, and **two** handlers (`get_architecture_overview`, `get_refactor_report`) serve it. It runs **only** on the null path, so an ordinary read pays nothing, and it re-derives rather than threading state through 56 call sites.

**Scope, stated honestly:** "No analysis found" appears about 36 times across 27 files, including one still in the very file this PR edits (`analysis.ts`, `handleGetSignatures`) and one in `architecture.ts`. This lands the mechanism and converts two call sites; the remaining ones are a deliberate follow-up, not an oversight. `classifyToolError` is untouched — no live regression there, noted as a follow-up.

The verdict is the repository's existing `NotReadyResult` shape — `error` + `notReady` + `reason` + `remedy` — so an agent branches on **one** `reason` taxonomy rather than a second parallel vocabulary. `NotReadyReason` gains four members:

| reason | what it means | remedy |
|---|---|---|
| `index-absent` | no index | `openlore analyze` — the original sentence, unchanged |
| `index-publish-in-progress` | a writer is mid-publish right now | **retry shortly** — expected, not damage |
| `index-generation-mismatch` | artifacts rewritten, manifest not republished, **and no writer holds the lock** | `openlore analyze`, and a publish was lost |
| `index-generation-unavailable` | manifest present and refused (malformed, oversized, not a regular file) | `openlore analyze` |
| `index-unreadable` | present and coherent, still not loadable — or an artifact the reader refused (symlink, FIFO) | `openlore analyze` |

The first-run case keeps the wording it always deserved. The other three now say that something went wrong, rather than that something was never set up.

## A correction the code made to me

The first draft called `generation-unavailable` "missing or unreadable", and tested it by **deleting** `generation.json`. That is wrong: an absent manifest is a legitimate legacy analysis, which `readCurrentGeneration` synthesizes a generation for, so it never reaches this diagnosis.

Both the message and the test now say "present but refused", and a fifth test asserts the legacy case explicitly — an absent manifest is not damage.

The diagnosis is deliberately shy elsewhere too: it answers `unreadable` rather than inventing a cause it did not observe.

## Hardening pass (second commit)

Review found three places where the *diagnosis* itself guessed — breaking the PR's own thesis that a confident wrong answer is worse than a vague one.

1. **It followed symlinks; the reader does not.** `loadPublished` opens `O_NOFOLLOW` and verifies the descriptor against the path entry, so a symlinked artifact is refused. A path-following `stat` reports the *target*: a symlink to a valid artifact was diagnosed `generation-mismatch`, a dangling one `absent` — the same lie, one layer up. Now `lstat`.
2. **It called a healthy publish damage.** `markGenerationUnavailable` deliberately writes a well-formed `{version, state:'publishing'}` manifest before the first replacement of a *normal* publish, and `readCurrentGeneration` returns null for it by design. A new `generationPublishInProgress` recognizes exactly that sentinel and reports a transient in-flight publish.
3. **It asserted an incident on the ordinary mid-write window.** A writer publishes the manifest last, so a mismatch is *expected* during any concurrent analyze. A new non-blocking `isAnalysisLockHeld` (the existing `isDecisionsLockHeld` peek, generalized) is consulted: a held lock says retry; only an unheld one claims a lost publish.

Plus: four handler-level tests (nothing pinned the handlers to the wiring before), and a spec delta next to `ReadyOrHonestFirstUse`.

Six of the fourteen tests fail against the pre-hardening logic with exactly the mislabels above, and pass after.

## Verification

- 14/14 tests in `index-unservable-diagnosis.test.ts`, including the exact observed shape (publish, rewrite the artifact, do not republish).
- Full unit suite: identical failing-file set to the same suite with this change stashed (17 pre-existing files); the two extra files in the branch run pass in isolation, i.e. load-induced timeouts.
- Lint, typecheck and build clean.

### DIFF
--- a/openspec/specs/mcp-handlers/spec.md
+++ b/openspec/specs/mcp-handlers/spec.md
@@ -1519,6 +1519,55 @@ an agent can decide to proceed on the disclosed-stale answer or retry.
 - **THEN** the responses respectively carry `reason: index-absent`, the staleness verdict
   with a repair-in-progress marker and reason, and no freshness caveat at all
 
+### Requirement: AnUnservableIndexNamesWhyRatherThanReportingAbsence
+
+A tool that cannot serve an index SHALL report WHY in the same machine-readable not-ready
+shape and the same `reason` taxonomy as `ReadyOrHonestFirstUse` — never a second, parallel
+vocabulary — and SHALL NOT report a present-but-refused index as an absent one. Beyond
+`index-absent` and `graph-unavailable`, the taxonomy SHALL distinguish at least: an artifact
+the reader refused as not a regular file, a generation manifest present and refused, a
+publish currently in flight, and artifacts that no longer match their published generation.
+
+The verdict SHALL be derived from what was OBSERVED, not from what a path-following read
+would infer:
+
+- The artifact's kind SHALL be determined without following links, because the production
+  reader refuses a symlinked artifact outright. A symlinked or dangling-symlinked artifact is
+  therefore refused, never reported as a generation mismatch or as an absent index.
+- The deliberate in-flight-publish manifest sentinel — the well-formed
+  `{version, state:'publishing'}` a writer commits before its first artifact replacement, for
+  which the manifest reader answers "unavailable" BY DESIGN — SHALL be reported as a transient
+  publish in progress whose remedy is to retry, never as a damaged publish.
+- An artifact/generation mismatch SHALL NOT be asserted as a lost publish while a writer is
+  observed inside the artifact-write critical section, because a writer rewrites artifacts in
+  place and publishes the manifest last, making a mismatch the EXPECTED state for the duration
+  of any concurrent write. The incident claim SHALL be conditional on no such writer being
+  observed.
+
+#### Scenario: A refused symlinked artifact is not reported as a mismatch or as absence
+
+- **GIVEN** an analysis whose `llm-context.json` is a symbolic link, whether to a valid
+  generation-matching artifact or to nothing at all
+- **WHEN** a tool that cannot serve it reports why
+- **THEN** the verdict names the artifact as not a regular file, and is neither
+  `index-absent` nor `index-generation-mismatch`
+
+#### Scenario: A publish in flight is transient, not damage
+
+- **GIVEN** a writer that has committed the in-flight-publish manifest sentinel and not yet
+  published the new generation
+- **WHEN** a tool that cannot serve the index reports why
+- **THEN** the verdict names a publish in progress, its remedy is to retry rather than to
+  rebuild, and it does not claim the publish is damaged
+
+#### Scenario: A mismatch is an incident only when no writer is running
+
+- **GIVEN** two repositories with byte-identical rewritten-but-unrepublished artifacts, one
+  with the artifact-write lock held by a live writer and one with no writer at all
+- **WHEN** each reports why the index is unservable
+- **THEN** the first names the expected mid-write window and asks the caller to retry, and
+  only the second claims a publish was lost
+
 ### Requirement: PriorChurnIsMeasuredBeforeTheBriefedRange
 
 The system SHALL compute `briefing_since`'s prior-churn evidence — `priorChurn`, its
--- a/src/core/runtime/advisory-lock.ts
+++ b/src/core/runtime/advisory-lock.ts
@@ -521,7 +521,25 @@ export async function withAnalysisLock<T>(analysisDir: string, fn: () => Promise
  * `record_decision` spawns against the run already underway.
  */
 export async function isDecisionsLockHeld(rootPath: string): Promise<boolean> {
-  const lockPath = join(decisionsDir(rootPath), DECISIONS_LOCK_FILE);
+  return isLockHeldAt(decisionsDir(rootPath), DECISIONS_LOCK_FILE);
+}
+
+/**
+ * Non-blocking check: is a writer currently inside the artifact-write critical section
+ * for this analysis directory?
+ *
+ * The one honest way to tell an ORDINARY mid-write window apart from a genuine incident.
+ * A writer rewrites artifacts in place and publishes the manifest LAST, so an
+ * artifact/manifest mismatch is the EXPECTED state while `analyze` or a watcher persist is
+ * running — and an incident only when nobody is writing. Never acquires, steals, or waits.
+ */
+export async function isAnalysisLockHeld(analysisDir: string): Promise<boolean> {
+  return isLockHeldAt(analysisDir, ANALYSIS_LOCK_FILE);
+}
+
+/** Shared body of the non-blocking lock peeks: present, a regular file, and not stale. */
+async function isLockHeldAt(dir: string, lockFile: string): Promise<boolean> {
+  const lockPath = join(dir, lockFile);
   try {
     const handle = await open(lockPath, constants.O_RDONLY | constants.O_NOFOLLOW);
     try {
--- a/src/core/runtime/analysis-generation.ts
+++ b/src/core/runtime/analysis-generation.ts
@@ -335,6 +335,29 @@ export async function discardGeneration(analysisDir: string): Promise<void> {
 export async function markGenerationUnavailable(analysisDir: string): Promise<void> {
   await atomicWriteFile(
     manifestPathOf(analysisDir),
-    JSON.stringify({ version: GENERATION_MANIFEST_VERSION, state: 'publishing' }),
+    JSON.stringify({ version: GENERATION_MANIFEST_VERSION, state: PUBLISHING_SENTINEL_STATE }),
   );
+}
+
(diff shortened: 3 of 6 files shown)


clay-good:
Thank you for this — genuinely. The diagnosis you found is real, the write-up is one of the clearest bug narratives this repo has had, and the correction you documented on yourself (the first draft testing `generation-unavailable` by *deleting* `generation.json`, when an absent manifest is a legitimate legacy analysis) is exactly the reasoning we want in this area. The framing that the message *is* the quiet downgrade `loadPartialFirstRun`'s docstring exists to prevent is what convinced me this belongs on main.

Review turned up three cases where the *diagnosis* itself guessed, which breaks your own thesis — a confident wrong answer being worse than a vague one. Rather than send it back, I've pushed a second commit to your branch (`354e3734`) so it can land. Nothing was reverted; everything below is additive to what you wrote. Details, so you can check my work:

**1. `stat` follows symlinks, but the production reader does not.**

`loadPublished` opens with `O_NOFOLLOW` plus `descriptorIsThePathEntry`, so it refuses a symlinked artifact outright — telemetry reason `artifact_not_a_regular_file`. A path-following `stat` reports the *target*, so a symlink to a perfectly valid, generation-matching artifact came back `generation-mismatch` ("a publish is being lost"), and a **dangling** symlink came back `absent`. That is the same lie the PR removes, restated one layer up. Now `lstat`, which describes the entry the reader actually refused, and the message names the kind it saw. (`stat` doesn't open the file, so the FIFO-blocking concern that forces the reader's `O_NONBLOCK` doesn't arise here.)

**2. `generation-unavailable` excluded its most likely producer.**

`markGenerationUnavailable` (`analysis-generation.ts`) *deliberately* writes a well-formed `{version, state:'publishing'}` manifest, and `readCurrentGeneration` returns null for it **by design** — that's the commit protocol, called at `index-bundle.ts` before the first replacement of a completely normal, healthy publish. So "this is a damaged publish" fired on the protocol working exactly as specified. A new `generationPublishInProgress` recognizes only that exact sentinel, and the verdict is now a transient `index-publish-in-progress` whose remedy is *retry*, not *rebuild*.

**3. `generation-mismatch` over-claimed on the ordinary mid-write window.**

You quoted the comment yourself: a writer updates artifacts in place and publishes the manifest **last**. That makes a mismatch the *expected* state for the whole duration of any concurrent `analyze` or watcher persist — so "a publish is being lost", hedged only by "if this recurs", asserts an incident from evidence that doesn't establish one. It now consults the artifact-write lock via a new non-blocking `isAnalysisLockHeld` (your existing `isDecisionsLockHeld` peek, generalized — no new locking mechanism). A held lock says "a writer is running, retry"; only an unheld lock claims a lost publish. Your five-minutes-with-nothing-writing observation is precisely the case that still reports the incident, which is the whole point.

**One taxonomy instead of two.** The `indexUnservable` field sat alongside the repo's already-spec'd `notReady` / `reason` / `remedy` contract, leaving an agent two vocabularies for one question. `diagnoseIndexUnservable` now returns a `NotReadyResult`, and `NotReadyReason` grows `index-publish-in-progress`, `index-generation-unavailable`, `index-generation-mismatch`, `index-unreadable`. `notReadyResult` takes a remedy so "retry shortly" isn't mislabelled "openlore analyze". A spec delta sits next to `ReadyOrHonestFirstUse` in `openspec/specs/mcp-handlers/spec.md` — this repo has doc-claim-sync guard tests, so observable tool-output semantics need to be written down.

**Tests.** Yours all survive (retargeted to the new reason names). Added: the three regression cases above, and four handler-level tests — nothing pinned `handleGetArchitectureOverview` / `handleGetRefactorReport` to the new wiring, so a refactor could have silently restored the flat sentence with every test still green. Six of the fourteen fail against the pre-hardening logic with exactly the predicted mislabels:

```
expected 'index-generation-mismatch'    to be 'index-unreadable'            (symlink to a valid artifact)
expected 'index-absent'                 to be 'index-unreadable'            (dangling symlink)
expected 'index-generation-unavailable' to be 'index-publish-in-progress'   (healthy in-flight publish)
expected 'index-generation-mismatch'    to be 'index-publish-in-progress'   (writer holds the lock)
```

**One correction to the PR body**, which I've also edited: it said "the two handlers that answered the old sentence", but "No analysis found" appears about 36 times across 27 files — including one still in the very file this PR edits (`analysis.ts`, `handleGetSignatures`) and one in `architecture.ts`. Landing two sites is fine; the body now says so rather than implying the conversion is complete. `classifyToolError` is a follow-up too — no live regression there, so I left it alone.

Lint, typecheck and build are clean. The full unit suite's failing-file set is identical with and without the change (17 pre-existing files); two further files that failed in the branch run pass in isolation, so they're load-induced timeouts, not regressions.

Not merging myself — leaving that to @clay-good. Thanks again for chasing this down from a hash comparison you did by hand.

clay-good (approved):
LGTM after the hardening pass.

The premise is right and worth having: `readCachedContext` already computes a diagnosis and throws it away at the `return null`, and re-deriving it only on the error path costs nothing on the happy path. Thanks @laurentftech — this is a real improvement to how the substrate answers when it can't serve.

Review found three cases where the first draft printed a *wrong* reason, which matters more than usual in a change whose whole thesis is honesty. All three are now fixed and pinned by tests:

- **Symlinks.** The diagnosis used a path-based `stat` while the production reader (`loadPublished`) refuses symlinks via `O_NOFOLLOW`. A symlink to a valid artifact reported `generation-mismatch` ("a publish is being lost"); a dangling one reported `absent` — exactly the lie this PR exists to remove. Now `lstat`, both reported as not-a-regular-file.
- **In-flight publish.** `markGenerationUnavailable` deliberately writes a well-formed `{version, state:'publishing'}` sentinel during every healthy import promotion. That was being reported as "a damaged publish". Now recognized as its own `index-publish-in-progress` state.
- **Mid-write window.** A generation mismatch is the *expected* state during any concurrent `analyze`, but the message asserted an incident. Now consults the artifact-write lock: held means retry, unheld is the genuine incident.

Also folded into the existing `NotReadyReason`/`notReadyResult` contract rather than adding a parallel `indexUnservable` field, so agents have one taxonomy to branch on, with a matching spec requirement alongside `ReadyOrHonestFirstUse`.

Before/after evidence is clean: reverting just the function body fails 6 tests with precisely the predicted mislabels. The lock pair is the strongest — byte-identical artifacts on disk, only the lock differs, and the verdict flips.

Scope is incremental by intent ("No analysis found" still appears ~36 times across 27 files) and the body now says so plainly. All 11 checks green including Windows.

PR #484, opened 2026-09-12
fix(install): quote the Windows hook path Git Bash was eating

### LINKED ISSUE #483: Windows Hook Path Quoting Bug
**What happened**
When running `openlore install` on Windows, the installer creates hook commands in `.claude/settings.json` and `~/.claude/settings.json` where the `node.exe` executable path is quoted, but the script path argument is left unquoted.

Because Claude Code uses Git Bash by default on Windows, Git Bash interprets the unquoted backslashes in the script path as escape characters and strips them out. This mangles the path into an invalid drive-relative string, causing every `SessionStart` and `UserPromptSubmit` hook execution to fail.

Expected behavior: The script path argument should be enclosed in quotes so Git Bash preserves the backslashes and resolves the module properly.

**The command you ran**

```
openlore install

```

**Error message or unexpected output**
Generated command in `.claude/settings.json`:

```json
"command": "\"C:\\Program Files\\nodejs\\node.exe\" C:\\Users\\\\AppData\\Roaming\\nvm\\v24.13.0\\node_modules\\openlore\\dist\\cli\\index.js orient --json"

```

Console output on every prompt invocation:

```
SessionStart:startup hook error
Failed with non-blocking status code: node:internal/modules/cjs/loader:1424
  throw err;
  ^

Error: Cannot find module 'C:\Users\\\UsersAppDataRoamingnvmv24.13.0node_modulesopenloredistcliindex.js'
    at Module._resolveFilename (node:internal/modules/cjs/loader:1421:15)

```

**Environment**

* OS: Windows 11 Pro

* Node.js version (`node --version`): v24.13.0 (via nvm-windows)

* OpenLore version (`openlore --version`): 3.1.1

**Anything else**

* **Steps to Reproduce:**
1. On Windows with Git Bash installed, run `openlore install` for the Claude Code adapter.

2. Inspect `.claude/settings.json` or `~/.claude/settings.json` to see the unquoted script path.

3. Start `claude` or send any prompt to trigger the `UserPromptSubmit` / `SessionStart` hook error.

* **Suggested Fix:**
Quote all path arguments when generating Windows hook commands (e.g., in `src/utils/platform-command.ts`):

```json
"command": "\"C:\\Program Files\\nodejs\\node.exe\" \"C:\\Users\\\\AppData\\Roaming\\nvm\\v24.13.0\\node_modules\\openlore\\dist\\cli\\index.js\" orient --json"

```

* **Workaround:**
Manually add double quotes around the script-path argument in `.claude/settings.json` and `~/.claude/settings.json`. Note that re-running `openlore install` will overwrite this fix.

L4XB:
## What & why

Fixes #483.

`formatPlatformCommand` quotes a Windows argument only when it contains a space or a cmd.exe metacharacter:

```ts
.map((part) => /[\s&|<>^%!()]/.test(part) ? `"${part}"` : part)
```

Claude Code runs a hook command through Git Bash on Windows, and there an unquoted backslash is an escape character that gets dropped. `C:\Program Files\nodejs\node.exe` was quoted because it has a space; the entry path was not, because an nvm-windows install has none. So `.claude/settings.json` got

```
"C:\\Program Files\\nodejs\\node.exe" C:\\Users\\me\\AppData\\Roaming\\nvm\\...\\index.js orient --json
```

and Git Bash resolved the second path as `C:Usersme...index.js`, which is the reporter's `Cannot find module` on every `SessionStart` and `UserPromptSubmit`.

A backslash now triggers the same double quoting. Both shells accept that form: double quotes are the only grouping cmd.exe understands, and Git Bash keeps the backslashes inside them. The POSIX branch is untouched, deliberately: `quotePosix` already single-quotes anything outside its allowlist, and widening the double-quote form to POSIX would reopen `$` and backtick expansion that the existing test at `platform-command.test.ts` pins.

Known limit, unchanged by this PR and not worth more machinery: inside double quotes Git Bash still collapses `\\`, `\$` and `` \` ``, so a UNC path or a directory literally named `$Recycle.Bin` stays mangled. No single string form is correct for both shells there.

## Checklist

- [x] `npm run typecheck` passes
- [x] `npm run lint` passes
- [ ] `npm run test:run` passes
- [x] Added/updated tests for the change
- [ ] If I touched `src/core/analyzer/`, `src/core/generator/stages/`, or `src/core/services/mcp-handlers/`: ran `npm run test:e2e` (not touched)
- [ ] If I added/changed an MCP tool or Pi-extension behavior: kept the two surfaces in parity (not touched)
- [ ] Updated docs (README / `docs/` / specs) if behavior or counts changed (no counts or documented behaviour changed)

On the third box, honestly: `npm run test:run` is red on a clean checkout of `main` at `9e6b297` here, 10 files and 65 tests, and it is red in exactly the same way with this branch. The suites this change can affect pass:

```
npm run test:run -- src/utils/platform-command.test.ts src/cli/install/adapters/platform-commands.test.ts src/workflow-security.test.ts
Test Files 3 passed (3), Tests 53 passed (53)
```

One note if you reproduce this: `npm install` rewrites `package-lock.json` and drops five integrity digests, which then fails `workflow-security.test.ts`. Restoring the lockfile clears it; the lockfile is not part of this PR.

## Notes for reviewers

The existing Windows case in `platform-command.test.ts` encoded the bug: it asserted the second path stays bare. It is renamed to say which shell the quoting is for and now expects it quoted, plus a second case with the reporter's own nvm-style path and the `orient --json` flags, which stay unquoted. Both fail if the backslash is removed from the character class.

## Maintainer follow-up (hardening pushed onto this branch)

**Status: LGTM on the diagnosis and the outcome.** @L4XB found a real bug, named the right
shell, and fixed the case users are hitting. Their commit is kept first; three commits on top
harden it. Thank you — this was a good report and a good patch.

### What was wrong, beyond the reported case

Quoting the entry path is correct. Getting there by adding `\` to a **denylist** is not: a
denylist is only as complete as the next character someone thinks of, and it only governs the
parts that get quoted. A part emitted **bare** is exposed to an entirely different set of
metacharacters. Over 900 random parts, **84** that the denylist left bare came back mangled or
dangerous from a real shell — `;`, `'`, `~`, `*`, `?`, `#`, brace expansion, the empty string.
`a;id` on its own runs `id`.

The POSIX branch never had this problem: `quotePosix` is an **allowlist** and single-quotes
everything else, sound by construction. The two branches had simply drifted.

Double quotes also do not make a POSIX shell literal, so four shapes still could not be
carried — two of them *worse* than the bug being fixed:

| Part | What the shell does |
|---|---|
| embedded `"` | ends the quoted run, so **the rest of the line executes** |
| trailing `\` | escapes our own closing quote, **swallowing later arguments** |
| `$…` / backtick | substituted, silently wrong |
| UNC prefix, `\$` | loses a backslash — this issue's `Cannot find module`, again |

### How it was fixed

- **One shared allowlist** for both branches, so neither can be weaker than the other for the
  same input. Every real path gains exactly the quotes this PR added; already-quoted forms are
  byte-identical.
- **The four unrepresentable shapes are refused**, not emitted. No string means the same thing
  to cmd.exe and to a POSIX shell for them.
- **A refusal costs the hooks and nothing else.** An unquotable path is a property of the host,
  not a clash with the user's file, so it is not a conflict: the MCP entry (an argv), the
  instruction block and the tool permission are all still written. A hook wired *earlier* is
  removed rather than left naming a command the host mangles; a user-authored entry in the same
  group survives.
- **Uninstall no longer formats a command at all** — it needs only the hook keys. Formatting
  there would have thrown and stranded the very hooks uninstall exists to remove.
- **`openlore update` prints the plain package-manager command on Windows.** The resolved form
  is not a command in PowerShell: a statement whose first token is a quoted path parses as a
  string expression.

### Proof it works

- **A real shell is the judge.** `platform-command.posix-oracle.test.ts` emits the command the
  installer would write, hands it to `bash`, and compares the returned argv. It pins the
  reporter's exact mangled path, proves the injection case executes when unguarded, and drives
  the emitter over a hostile corpus — which reports **84 failures against the denylist and 0
  against the allowlist**. This is the test class whose absence let the bug ship green: every
  previous unit test restated the code's own assumption about a shell.
- **A real Windows host checks the real hook.** The Windows smoke job now writes the emitted
  hook command to a script, runs it under the runner's own Git Bash, and requires a version
  back. It passed — so the fix is confirmed on Windows + Git Bash, not just reasoned about.
- The guard models bash's documented double-quote rule and was reconciled against a real
  `bash` over 40,000 inputs, in both directions (no false accepts, no false refusals).

### Notes

- `bash`'s deprecated `$[1+1]` arithmetic is still evaluated, so `[` belongs in the
  expansion-start set. The original fuzz alphabet had no `[` and structurally could not find it.
- A `$` before a separator, a space, or end of string is inert, so `C:\Users\dev$\…` stays
  installable rather than being refused out of caution.
- Known unsupported hosts, stated rather than papered over: a **UNC** install and a directory
  whose name is a shell expansion get everything except the hooks, with the reason named.
  cmd.exe's `%VAR%` and `!VAR!` expansion is undefendable by any quoting form.
- On the red suite: `main` is not red. The failures reproduce as stale `node_modules` plus a
  missing `npm run build` — CI runs `npm ci` + `npm run build` + `test:equivalence` +
  `test:unit`, never `test:run`, and the hook-installation tests need `dist/`. Details in a
  comment below.
- Two adjacent crashes are **not** addressed here, deliberately, because they are pre-existing
  and have a different cause — cannot *locate* a command, rather than cannot *quote* one.
  `resolveOpenloreCommand` still throws out of a project-scope adapter when `npx-cli.js` cannot
  be found or the Node path is not absolute (`mcpEntry` runs before any check), and
  `doctor --fix` discards `runInstall`'s exit code and reports a rewire as corrected either way.
  Both are filed as follow-ups; fixing them here would widen the claim this PR makes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -322,6 +322,30 @@ jobs:
               throw "Windows install routed the $key hook through npx (cmd.exe shim)"
             }
           }
+          # harden-windows-hook-quoting (#483): Claude Code runs a hook command through GIT
+          # BASH, where a bare backslash is an escape that is dropped. Nothing above can see
+          # that - the MCP launcher is spawned with an argv and never meets a shell - so the
+          # unquoted entry path shipped green and every hook died with `Cannot find module`.
+          # This runs the hook STRING through the real Git Bash on a real Windows host, which
+          # is the only place in CI where that interpretation is the genuine article.
+          $bash = (Get-Command bash -ErrorAction SilentlyContinue).Source
+          if (-not $bash) { throw 'Git Bash is absent from the runner; the hook shell cannot be verified' }
+          foreach ($key in 'SessionStart', 'UserPromptSubmit') {
+            # Swap the subcommand for `--version`, so the probe proves the SHELL resolved the
+            # path and Node ran it, without depending on an index being built yet.
+            $probe = $settings.hooks.$key[0].hooks[0].command -replace 'orient (--json|--inject)$', '--version'
+            # Written to a file, not passed as an argument: the hook string carries its own
+            # double quotes, and PowerShell's native-argument quoting would rewrite them.
+            Set-Content -Path hook-probe.sh -Value $probe -Encoding ascii
+            $out = (& $bash hook-probe.sh 2>&1) | Out-String
+            if ($out -match 'Cannot find module') {
+              throw "#483 regression: Git Bash could not resolve the $key hook path. $out"
+            }
+            if ($out -notmatch '\d+\.\d+\.\d+') {
+              throw "The $key hook command did not run under Git Bash. $out"
+            }
+          }
+          Remove-Item hook-probe.sh
           # A hostile npx.cmd in the working directory must now be irrelevant: the launcher
           # never reaches npx at all. Kept as a regression probe on exactly that property.
           Set-Content -Path npx.cmd -Value '@exit /b 91'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -336,7 +336,10 @@ jobs:
             $probe = $settings.hooks.$key[0].hooks[0].command -replace 'orient (--json|--inject)$', '--version'
             # Written to a file, not passed as an argument: the hook string carries its own
             # double quotes, and PowerShell's native-argument quoting would rewrite them.
-            Set-Content -Path hook-probe.sh -Value $probe -Encoding ascii
+            # utf8NoBOM, not ascii: an ascii write would silently mangle a non-ASCII path
+            # (`C:\Users\Müller\…`) before bash ever saw it, and a BOM would break the first
+            # token. The runner's own paths are ASCII, so this is about not lying if that changes.
+            Set-Content -Path hook-probe.sh -Value $probe -Encoding utf8NoBOM
             $out = (& $bash hook-probe.sh 2>&1) | Out-String
             if ($out -match 'Cannot find module') {
               throw "#483 regression: Git Bash could not resolve the $key hook path. $out"
--- a/openspec/changes/harden-windows-hook-quoting/.openspec.yaml
+++ b/openspec/changes/harden-windows-hook-quoting/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-09-12
--- a/openspec/changes/harden-windows-hook-quoting/proposal.md
+++ b/openspec/changes/harden-windows-hook-quoting/proposal.md
@@ -49,13 +49,25 @@ thing to cmd.exe and to a POSIX shell for them, so `formatPlatformCommand` throw
 writing a line that silently fails or executes code — the protection `quotePosix` already
 gives the POSIX branch, which the Windows branch was missing for the same input.
 
-**The refusal reaches the user as a refusal, not a crash.** `runInstall` rethrows a
-project-scope adapter error, so both config writers check first (`windowsCommandHazard`) and
-decline one file with the reason. `openlore update` only ever PRINTS a command — it spawns its
-own argv without a shell — so it degrades to the plain package-manager instruction for the
-detected method. That instruction is derived from the method, not from the resolved
-invocation: joining the resolved parts would have printed two unquoted absolute paths, which
-is less runnable than the line it replaces.
+**A refusal costs the hooks, and nothing else.** `runInstall` rethrows a project-scope adapter
+error, so the writers check first (`windowsCommandHazard`) instead of letting the throw escape.
+What they then do matters as much: an unquotable path is a property of the HOST, not a clash
+with the user's file, so it is not reported as a conflict. The MCP entry (an argv), the
+instruction block (prose) and the tool permission (a literal) are all still correct and are
+still written; only the hooks are left out. Treating it as a conflict returned exit 1, which
+also skipped the index build — one unwritable field cost the entire install.
+
+A hook OpenLore wired earlier is REMOVED on such a host, not left in place. It names a command
+this host mangles, so keeping it is #483's `Cannot find module` once per turn, forever, with
+nothing saying why; a user-authored entry in the same group survives untouched.
+
+`openlore update` only ever PRINTS a command — it spawns its own argv without a shell — so on
+Windows it prints the plain package-manager command for the detected method instead of the
+resolved form. That is not merely a fallback: a statement whose first token is a quoted path
+parses in PowerShell as a string EXPRESSION, so the resolved line is a syntax error there
+rather than an invocation. Printing the command the user actually meant is correct for cmd.exe,
+PowerShell and bash alike, and it keeps every `$`-shaped path out of a shell whose expansion
+rules differ from the POSIX ones the refusal models.
 
 **Uninstall is decoupled from formatting.** `managedHooks` was called on the removal path for
 its keys alone, so an unformattable path would have thrown there too — stranding the hooks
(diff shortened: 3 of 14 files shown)


clay-good:
Thanks @L4XB — good report, good patch, and the right diagnosis. You identified the shell
correctly, which is the hard part here: the Windows branch had been written for cmd.exe and
nothing in the repo said so out loud. I've pushed three commits on top of yours rather than
replacing it; your commit stays first and the details are in the description.

Two things that may be useful to you directly.

**On `npm run test:run` being red — `main` is fine, and your branch never broke it.** The
failures are environmental, and the mechanism is worth knowing for next time:

- CI never runs `test:run`. It runs `npm ci`, then **`npm run build`**, then
  `npm run test:equivalence` and `npm run test:unit` (together the same file set).
- The `npm run build` step is load-bearing for tests, not just for shipping: the
  hook-installation tests pin the same *built* external launcher that a real install wires, so
  without `dist/` present `resolveTrustedHookLauncher()` returns null and every hook test fails
  on any platform. That is the most likely source of a large cluster on a clean checkout.
- The rest reproduce as stale `node_modules` — a pinned dependency version differing from the
  installed one shows up as a plain assertion mismatch, which reads like a real failure.

So the closest local mirror of CI is:

```bash
npm ci && npm run build && npm run test:equivalence && npm run test:unit
```

**Your lockfile observation is correct and worth recording.** `npm install` really can drop
integrity digests (a nested shrinkwrap pins tarballs without an `integrity` field, and a
re-resolve copies them verbatim), which then trips the SRI assertion. `npm ci` avoids it
entirely, and `npm run lock:integrity:check` reports the state if you ever need to confirm.
Nothing for you to fix — you were right that it is not part of this PR.

One finding from your patch that is worth naming, because it is a genuinely subtle one: the
character class governs only the parts that get **quoted**. A part emitted bare is exposed to a
different set of metacharacters, and that is what the follow-up commits close by inverting the
rule to the allowlist the POSIX branch already used. Your change made the reported case correct;
the inversion makes the branch sound for the cases nobody had reported yet.

L4XB:
Thanks, and the CI recipe was the missing piece. I ran it against your pushed head and against `main` on the same machine, which settles the question your first bullet raises.

```
npm ci && npm run build && npm run test:equivalence && npm run test:unit
```

| suite | `main` | #484 head (your three commits on mine) |
|---|---|---|
| `test:equivalence` | 5 failed / 397 passed, 1 error | 5 failed / 397 passed, 1 error |
| `test:unit` | 20 failed / 9361 passed | 5 failed / 9441 passed |

So the head is not worse than `main` here, and the `build` step is exactly as load-bearing as you say: the Dart and Lua grammar cases that fail on my `main` run pass on the head run, where `npm ci` preceded the build.

What is left on the head splits into two environmental groups, neither touching installation:

- **Worker threads and WASM grammars**: the whole `extraction-pool-threads` file plus `analyze-pass-reuse`'s real-worker-lane case. This machine is shared with other work, and those are the cases that care.
- **Wall-clock budgets**: `atomic-store`'s two concurrent-CAS cases hit `Test timed out in 30000ms`, and the post-cap probe case took 64.9s to conclude. Both are time-bounded rather than assertion-bounded, so they report the host, not the code.

`progress.test.ts`'s token-count case is the only assertion failure I cannot attribute, and it fails identically on `main`, so it is not from this branch either.

Nothing in the hook-installation set failed on the head, which is the part I would have worried about after inverting the rule.

On that inversion: your reading is right and mine was narrower than I thought. I checked what the character class governed and concluded "the quoted parts are safe", without following where a part is emitted bare — and that is a different metacharacter set with a different answer. An allowlist has the property my denylist did not: a shape nobody has reported yet is refused rather than let through, and the POSIX branch was already the evidence that the shape works. Matching the two branches also removes the thing that made the Windows branch quietly special in the first place, which is what the report was really about.

The lockfile note is recorded on my side too: `npm ci` for anything I intend to compare against CI, and `npm run lock:integrity:check` when a digest question comes up.

